PR #202, opened 2017-06-20
Hebrew version

talargoni:
need to translate more sections,
will try to get help from OWASP Israel members


talargoni:
Hebrew version

danielcuthbert:
This is a great translation effort, thank you

PR #243, opened 2017-12-01
New requirement for directly setting user password

stuartgunter:
Added `2.34` to capture the requirement that a user be able to choose their own password and directly set it in the application without needing to reveal it to another party. The objective of this requirement is to avoid situations where users are given a password that has been set by another party (e.g. administrator) and are not given the ability to change it.


stuartgunter:
This PR originated from a discussion in Slack. Please see the conversation there for more context. The discussion started with this message: https://owasp.slack.com/archives/C06MNF14M/p1512067111000252

PR #303, opened 2018-11-02
V1 - Added "rudimentary threat analysis"

philippederyck:
Level 2 applications can benefit from a rudimentary threat model. The old 1.6 (current 1.7) addressed threat modeling for level 3, but doing nothing at all in this regard for level 2 makes little sense compared to the other requirements in the ASVS.


philippederyck:
This should resolve it (I'm working on my git wizardry :)). I'll have some more PRs heading your way over the coming days. Let me know if you prefer bigger ones instead of "per item PRs".

PR #323, opened 2018-12-29
Update 0x02-Preface.md - how it talks about OT10

tghosth:
I made some changes to the wording based on how I think it makes sense to refer to the OWASP Top 10 Risks project (there are so many OWASP Top 10s nowadays...)


danielcuthbert:
love this!!

PR #325, opened 2018-12-29
Update 0x03-Using-ASVS.md - Checked instead of Ensured

tghosth:
Checked instead of Ensured


tghosth:
@danielcuthbert did you intend to approve but close without merging?

PR #327, opened 2018-12-29
Update 0x24-V19-Config.md - hyphens to clarify an expression

tghosth:
Added hyphens to clarify an expression


tghosth:
Also made some wording changes

PR #341, opened 2018-12-29
Update 0x20-V15-BusLogic.md - A bit of a revamp

tghosth:
Starting making some significant changes to V15 as I think it needs to be a little more granular.


danielcuthbert:
this pull is solid, wow!

PR #354, opened 2019-01-03
Reorganise V4 - Ready for review

tghosth:
I have started to reorganise V4 as it is currently one long list and also seems to be missing concepts from the Proactive Controls. 

I will leave renumbering to the end.

### DIFF
--- a/4.0/en/0x12-V4-Access-Control.md
+++ b/4.0/en/0x12-V4-Access-Control.md
@@ -10,21 +10,38 @@ Authorization is the concept of allowing access to resources only to those permi
 
 ## Security Verification Requirements
 
+### V4.1 General Access Control design
+
 | # | Description | L1 | L2 | L3 | Since |
 | --- | --- | --- | --- | -- | -- |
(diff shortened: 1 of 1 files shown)


tghosth:
Final list of added controls from Proactive Controls. I think this is ready for review. If there are no objections after a few days I will merge.

| # | Description | L1 | L2 | L3 | Since |
| --- | --- | --- | --- | -- | -- |
| **4.1.1** | Verify that the chosen access control solution is flexible enough to meet the application's needs.  | ✓ | ✓ | ✓ | 4.0 |
| **4.1.6** | Verify that the principle of deny by default exists whereby new users/roles start with minimal or no permissions and users/roles do not receive access to new features until access is explicitly assigned.  | ✓ | ✓ | ✓ | 4.0 |
| **4.2.2** | Verify that attribute or feature-based access control is used whereby the code checks the user's authorization for a feature/data item rather than just their role. Permissions should still be allocated using roles. | ✓ | ✓ | ✓ | 4.0 |

danielcuthbert:
This was heavily needed and is actually needed all over. There was talk about doing a reset of the numbering, see https://github.com/OWASP/ASVS/issues/298

I'm good with this pull request and will merge, means we can crack on and do the rest. Thanks

PR #358, opened 2019-01-08
Rework V8 to include Error Handling controls - IN PROGRESS

tghosth:
This section talks about logging but not handling where I think it is important to include a few controls as well.

I am in the middle of working on this.

### DIFF
--- a/4.0/en/0x15-V8-Error-Logging.md
+++ b/4.0/en/0x15-V8-Error-Logging.md
@@ -20,17 +20,17 @@ It is also important to ensure that the application fails securely and that erro
 
 | # | Description | L1 | L2 | L3 | Since |
 | --- | --- | --- | --- | -- | -- |
-| **8.1** | Verify that the application does not log credentials, session tokens or payment details. | ✓ | ✓ | ✓ | 4.0 |
-| **8.2** | Verify that the application does not log other sensitive data as defined under local privacy laws or relevant security policy. |  | ✓ | ✓ | 3.0 |
-| **8.3** | Verify that the application logs security relevant events including successful and failed authentication events, access control failures and input validation failures. | ✓ | ✓ | ✓ | 4.0 |
-| **8.4** | Verify that each log event includes necessary information that would allow for a detailed investigation of the timeline when an event happens. |  | ✓ | ✓ | 1.0 |
-| **8.5** | Verify that all events are protected from injection when viewed in log viewing software. |  | ✓ | ✓ | 1.0 |
-| **8.6** | Verify that security logs are protected from unauthorized access and modification. |  | ✓ | ✓ | 1.0 |
-| **8.7** | Verify that the application appropriately encodes user-supplied data to prevent log injection. | ✓ | ✓ | ✓ | 4.0 |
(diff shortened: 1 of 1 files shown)


4.0/en/0x15-V8-Error-Logging.md
```diff
@@ -28,6 +32,14 @@ If logs contain private or sensitive data, the definition of which varies from c
 | **8.10** | Verify that all authentication decisions are logged, without storing sensitive session identifiers or memorized secrets. This should include requests with relevant metadata needed for security investigations.  | ✓ | ✓ | ✓ | 4.0 |
 | **8.11** | Verify that a common logging format and approach is used across the system.  | ✓ | ✓ | ✓ | 4.0 |
 
+### 8.2 Error Handling
+
+| # | Description | L1 | L2 | L3 | Since |
+| --- | --- | --- | --- | -- | -- |
+| **8.2.1** | Verify that a generic message is shown when an error occurs, potentially with a unique ID which support personnel can use to investigate.  | ✓ | ✓ | ✓ | 4.0 |
+| **8.3.2** | Verify that exception handling is used across the codebase to account for expected or unexpected error conditions.  | ✓ | ✓ | ✓ | 4.0 |
```
mastacheata: Shouldn't this be expected AND unexpected error conditions?
Otherwise this could be understood to either handle all known exceptions (i.e. catch only exceptions you throw yourself) or to have a fallback exception handling (i.e. a generic exception handler) and skip properly handling known exceptions intentionally thrown by application code.

mastacheata: Basically that's what 8.3.3 is for, but I still feel like this rule is easy to be misunderstood.

tghosth: Fair one, updated.

tghosth:
@mastacheata do you have any ideas for other requirements to add here?

mastacheata:
Not exactly sure.
I've learned to build errors in a way that the message contains context and is actionable.
(i.e.: for Form validation errors, you should tell the user which field has the wrong info and what the expected format is)
There's obviously one exception to that rule when it comes to authentication errors, which are supposed to be a generic authentication failed in order to not give hints for brute force / dictionary attacks.
But making ALL error messages generic seems a little over the top to me. Even from a security perspective.

4.0/en/0x15-V8-Error-Logging.md
mastacheata: Just noticed: You go from 8.2.1 to 8.3.2 and 8.3.3.
Also - but this might be on your roadmap anyways - you added a 3rd numbering component on 8.2, but haven't updated the 8.1 entries to use a middle component as well.

tghosth: Yeah numbering is on the list of things to do at the end :)

tghosth:
Thanks @mastacheata, I softened the generic error messages requirement [8.2.1](https://github.com/OWASP/ASVS/blob/b6d010d7d97508af57dc8766de4a7e2107cc3930/4.0/en/0x15-V8-Error-Logging.md#82-error-handling)

Issue #360, opened 2019-01-08
Should ASVS include guidelines to mitigate effects of a security breach? (Backups etc.)


mastacheata:
The ASVS already includes some guidelines that would only take effect if a security breach or mishap has happened before, like not storing plain text passwords in the sourcecode. (Users shouldn't be able to see the sourcecode anyway, so this is a non-issue unless you already had a security problem)

I was wondering if backups should be included in the scope of the ASVS as well?

tghosth:
Sounds like it should be in [V9](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x16-V9-Data-Protection.md). Fancy creating a PR with some calls?

tghosth:
@mastacheata can this now be closed?

mastacheata:
Unless anyone got other ideas about this more generic suggestion, sure.
Backups is the only thing that comes to my mind that isn't already covered as a security thing (i.e. protecting personal data, not storing plaintext passwords, having logs of all authentication decisions etc.)

tghosth:
Ok so I think I will close for now, thanks for your help :)
If you have any other ideas, feel free to open a PR!

PR #363, opened 2019-01-08
Draft: Add backups to 0x24-V19-Config.md

mastacheata:
I think this is security relevant because an app that is not available is by definition insecure.
Also the ASVS already includes guidelines for things that should not be relevant during "normal" operation like not including plaintext passwords in the sourcecode. (which users should never have access to anyway - obviously except for open source projects, but there are other rules not applicable to open source as well)

I listed this under 19.5 because it's a more generic concern. Not sure if this would fit under 19.1 or if it's worth establishing an own subheading for this.

Regarding which parts should be covered by a backup, I'd love to hear feedback. Not sure if this sentence I wrote there is getting too convoluted by listing the things to backup. I'm not really happy with the "other custom components", but couldn't think of a better description at this very moment. Maybe a native speaker has a better idea for wording this?

Last but not least: It's super important to not only perform backups, but to also check if you can restore from a backup. That's probably the biggest problem with backups: Even those people who perform regular backups barely ever check if they actually include all the necessary things and if there is a proper process for restoring from a backup. It's too late to check that when you NEED to restore from a backup.


tghosth:
This is what you talked about in #360, right? Would it not be better in section 9 which talks about data protection?

tghosth:
> This is what you talked about in #360, right? Would it not be better in section 9 which talks about data protection?

Hey @mastacheata, any thoughts on this?

mastacheata:
From the headline, V09 would fit too, but I understand V09 to be about personal or otherwise sensitive data.
Backups are a more generic thing, that's why I put it into V19.

tghosth:
Thanks @mastacheata, I removed the customer data part and I think I will add that separately to V9.

tghosth:
See commit to V9 [here](https://github.com/OWASP/ASVS/commit/4d3fdaf9970c8597d02ce0f33dec81b798713f48#diff-9ccb5e4dc576e32311b114badd1f1dc0).

PR #364, opened 2019-01-08
Add newlines for project leads

mastacheata:
Apparently the Project Leads were meant to appear on one line each, but someone forgot that markdown transforms single line-breaks to spaces unless preceded by 2 spaces themselves.


tghosth:
Makes sense to me

Issue #357, opened 2019-01-06
Final 4.0 release checklist


tghosth:
As I go along, I am noticing things that we probably only want to fix just before release so I will note them here, feel free to add :)

|Task|Assignee|Status|
 | --- | --- | -- |
|"_Pass over the entire thing to weed out useless controls and those that are un-developable or un-testable or are compound tests. An ASVS item should be atomic and not compound. It should be specific as to an action for (dev, testers, ops, security folks, auditors)_" - Andrew| | Complete |
|Full check of numbering (as requested in #298)| | Complete |
|Check heading formatting consistency| | Complete |
|Add CWE numbers (as requested in #67) | | Complete |
|<del>Check requirement number formatting consistency| @tghosth | Complete |
|Check page layout consistency| | Complete |
|Check spelling and grammar| | Complete |
|Check that all Top 10 Risks are L2, assuming that ratings are based on [this comment](https://github.com/OWASP/ASVS/issues/350#issuecomment-453808135). | | Complete |
|Check that all Top Ten Proactive controls are L2, assuming that ratings are based on [this comment](https://github.com/OWASP/ASVS/issues/350#issuecomment-453808135), see #322 for mapping. | | Complete |
|Move the IoT controls to an Appendix as per [this comment](https://github.com/OWASP/ASVS/pull/369#issuecomment-455332384). | | Complete |

mastacheata:
> TBD - Need to re-do case studies / applicability.

In [Using the ASVS](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x03-Using-ASVS.md)

vanderaj:
Suggest we leave Proactive Controls for now - many ended up in L1 due to being pen-testable. I'm going to mark them Complete.

vanderaj:
Release candidate 4.0 released. I created and paginated 4.0 PDF as well.

tghosth:
That L2 thing was when we were thinking about a different level rating method so I think that is fine now.

Amazing work getting this all pushed through!

PR #369, opened 2019-01-12
Remove the notion of physical tamper or epoxy

tkisason:
Hello everyone,

I would suggest we change the statement 20.18 to say:

* Verify the presence of physical tamper resistance and/or tamper detection features, including epoxy.
into:
* Verify the presence of tamper resistance and/or tamper detection features.

Tamper evidence and tamper resistance are something like security by obscurity, is it useful, yes, should it be used? (yes in addition to other controls and in very specific use-cases)

Here are some arguments why i changed this:
1) If we look at a run off the mill IoT device, what's the possible most likely threat model?
 - Attacker buys X devices, reverse engineers them to find juicy stuff (shared keys, certs, vulnerable SW or vulns in SW, etc) - Tamper evidence or resistance is moot here. Attacker will shave off epoxy, x-ray the device, etc, etc, and won't care for tamper detection. 
 - Attacker has access to a device, for let's say x hours, but needs to return that device to the owner (think interdiction attacks, covert physical or black bag attacks) - This is a specific issue where tamper evidence is welcome, but in the real world it is not really useful or will be useful. Epoxy is a *extremely* dangerous notion, because people think it's the solution to prevent tampering instead of implementing proper secure boot, trusted execution and encrypted storage on modern SoCs devices.

Epoxy as tamper evidence can be easily removed, and even more so, re-applied (in the case of an interdiction attack), so as a tamper-evident measure, it is of dubious value. If you want tamper detection it needs to be implemented like in modern HSMs or POS terminals, that means you implement something like a security mesh around the SoC / Memory, where it will be a physical-hardware method that will wipe the keys from the device or brick the device as a method to prevent extraction or tampering. But for modern SoCs that is not needed since you already have a secure enclave that if used right, makes those attacks very unfeasible.  

Regarding that, i would also consider upping this requirement to L3. Reason - Proper tamper detection is hard, you should implement trusted execution before that (I'll also send some suggestions in pull requests to other points regarding that). In general it is needed only for most critical stuff (PoS terminals, Medical, etc gear)...

Sincerely,
Tonimir


danielcuthbert:
I fully agree with everything here, you are spot on. I know we need to decide on what we are doing with the IoT section. Do we keep it and reference https://www.owasp.org/index.php/OWASP_Internet_of_Things_Project, or keep it and run the risk that the ASVS standard tries to include all, thus potentially making it a monster?

We do need to make this decision soon imho. @vanderaj & @jmanico thunks?

tkisason:
If you ask me, i would leave it here, until a more mature ASVS for IoT doesn't come along (IoTSVS?). The same thing happened with mobile. ASVS in start contained some baseline tips for mobile applications, afterwards MASVS became a mature project, and that's it, now ASVS references MASVS if you need mobile verification. I say a 1-2 pager for IoT in a very popular document is way better than a 1-2 pager on the wiki that might not get referenced.

jmanico:
Since IoT is out the gate now with the IoT T10 and similar; I’m ok with removing it and pointing to the IoT OWASP project....

--
Jim Manico
@Manicode
Secure Coding Education
+1 (808) 652-3805

tkisason:
Sure, but still IoT Top 10 shows top 10 vulnerabilities, and actually doesn't give a succinct guide on what's needed to implement a secure software product that also has a IoT/Embedded component? Those are 2 different things.

jmanico:
Yup. Just saying we have •something• to point to that is recent if we want to.

What about putting the IoT requirements in an appendix so it’s doesn’t mix with WebSec requirements?

Aloha,
--
Jim Manico
@Manicode
Secure Coding Education
+1 (808) 652-3805

tghosth:
I am going to merge this PR and I have made a note in #357 (the final 4.0 release checklist) to move this section to an appendix per Jim's [comment above](https://github.com/OWASP/ASVS/pull/369#issuecomment-455332384).

PR #371, opened 2019-01-12
Updated 20.7

tkisason:
Hello everyone,

I suggest we change the 20.7	statement of:
* Verify that ARM based processors make use of TrustZone and TEE (Trusted Execution Environment)

to:

* Verify that trusted execution is implemented and enabled, if available on the device SoC or CPU.

Reason is very simple, there are multiple TE methods available on multiple CPUs and SoCs so a general statement is better that mandates for anything that's non-ARM specific.

Sincerely,
Tonimir


tghosth:
This seems sensible and no-one has objected so I am going to merge it

PR #372, opened 2019-01-12
Removed 20.19

tkisason:
Hello everyone, 

I suggest that we remove: "20.19 Verify that identifying markings on chips have been removed"

The reason is simple -> It raises the manufacturing price, while providing only security by obscurity. Anyone that will decap the chip, can compare it to a known database like https://siliconpr0n.org/ , and if you want to extract data from a Flash chip, a emmc chip, it's pretty obvious what you need to look for.  Also, if you implement everything from L1, this raises the price of manufacturing and provides no additional security benefit except obscurity. 

Sincerely,
Tonimir


danielcuthbert:
Totally agree with this, it's not feasible at all and the cost of doing it is huge.

Issue #284, opened 2018-08-07
v4 Feedback: 0x21-V16-Files-Resources.md


ossie-git:
"Obtained from untrusted sources are stored outside the web root and limited permissions." -> "Untrusted file data obtained from untrusted sources are stored outside the web root and limited permissions"

16.3 Adding the need for AV I think is excessive as using a mime-aware library such as Apache Tika should largely remove the need for AV unless you allow users to upload dangerous file types such as EXEs or files that support active content (PDFs, Office documents, etc.)

There is no mention of the following controls:

- size limits + number of files uploaded limits
- quota limits per user (if needed)
- possibly using a different sub-domain to serve user uploads to prevent XSS
- the uploading of in-intentionally uploading executable code such as SVG images which are then served directly back to users

vanderaj:
Agreed

ossie-git:
I think that adding an item explicitly related to more thorough file type verification might be useful.  16.3 and 16.4 kind of touch on it but I think that we might want to emphasize two parts:

- not relying simply on file extensions and content-type sent (libmagic is useful).  
- verifying that the web server is configured correctly to not serve dynamic content (refer to the recent jQuery Plugin vuln (https://zd.net/2CUnToG) as an example)

The first won't completely remove issues such as polyglot attacks but when combined with the 2nd, should provide adequate protection

danielcuthbert:
I wonder if we need to look at the whole supply-chain issue here?

MazeOfThoughts:
Why did 16.4 get changed from 

> [...] to prevent remote/local file inclusion vulnerabilities.
to
> [...] to prevent remote/local code execution vulnerabilities.

This seems like a wholly different thing, where both of them are relevant. Wouldn't it make sense to have both of them as separate points?

jmanico:
This is a duplicate, closing and moving to #377

Issue #142, opened 2017-01-18
V16: Files and resources verification requirements comments / enhancements


vanderaj:
I found no mention of DoS type attacks (uploading very big files or a lot of very small files).  The only mention of this was in 18.4 which is specific to Web services.  I think this should be added.

16.5 CORS was mentioned here but just accepting untrusted data was mentioned.  I don’t think that there was anywhere in the standard where it mentioned securing CORS only to trusted domains.

16.7 - I think this can be expanded to also allow access to certain extensions so backup files, temporary files, etc are not accessible.  Or this can be added to V19

16.8 - I think this can be expanded to cover scenarios where malicious HTML can be uploaded to then used to execute XSS attacks as in this case, the browser’s client would execute it (not the application)

ossie-git:
Adding myself for follow-up

danielcuthbert:
Worry a little about the interpretation of these, they are good additions but a bit too open for my liking. We should also add here about 3rd-party CSS and others, especially in light of the recent supply-chain attacks.

tghosth:
Opened #377 to address these points, @danielcuthbert please can you see if you think this covers them all off.

PR #377, opened 2019-01-20
Updates to address issue #142 and #284 - READY FOR REVIEW

tghosth:
Here are responses to the points raised in #142

> I found no mention of DoS type attacks (uploading very big files or a lot of very small files). The only mention of this was in 18.4 which is specific to Web services. I think this should be added.

See: https://github.com/OWASP/ASVS/commit/88b77b5ed52bc04e6ae249771d707b9b3adf3cfb and https://github.com/OWASP/ASVS/pull/377/commits/6d3aea200d88ba818adf0d8777258775d3e1adb6

> 16.5 CORS was mentioned here but just accepting untrusted data was mentioned. I don’t think that there was anywhere in the standard where it mentioned securing CORS only to trusted domains.

See: https://github.com/OWASP/ASVS/commit/2a42ecd6b8716b70b71bab4a8547b7dd9e9e3b7d and https://github.com/OWASP/ASVS/pull/377/commits/69c4675786e2c4cf28cf992e640ac4b9aa3d3124

> 16.7 - I think this can be expanded to also allow access to certain extensions so backup files, temporary files, etc are not accessible. Or this can be added to V19

I think this is now covered by [16.10](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x21-V16-Files-Resources.md#references) `Verify that the web tier is configured to serve only files with specific file extensions to prevent unintentional information and source code leakage. For example, .bak, .swp and and similar extensions commonly used by editors should not be served by the web tier.`

> 16.8 - I think this can be expanded to cover scenarios where malicious HTML can be uploaded to then used to execute XSS attacks as in this case, the browser’s client would execute it (not the application)

See; https://github.com/OWASP/ASVS/pull/377/commits/0ed6ac26c6504a5a4182bd9cae506243d701938c

> IF this is the right section, I'd like to reference more of https://www.owasp.org/index.php/3rd_Party_Javascript_Management_Cheat_Sheet

So we now have [19.3](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x24-V19-Config.md#193-dependency) which includes various 3rd party dependency topics such as Subresource Integrity and updating and [19.4.7](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x24-V19-Config.md#194-http-configuration) which references Content Security Policy. Do you think that is sufficient?

### DIFF
--- a/4.0/en/0x20-V15-BusLogic.md
+++ b/4.0/en/0x20-V15-BusLogic.md
@@ -18,7 +18,7 @@ Ensure that a verified application satisfies the following high level requiremen
 | **15.4** | Verify the application monitors for unusual events or activity from a business logic perspective. For example, attempts to perform actions out of order or actions which a normal user would never attempt. |  | ✓ | ✓ | 2.0 |
 | **15.5** | Verify the application has configurable alerting when automated attacks or unusual activity is detected. |  | ✓ | ✓ | 2.0 |
 | **15.6** | Verify the application has automated reactions when automated attacks or unusual activity is detected. |  |  | ✓ | 2.0 |
-| **15.7** | Verify the application has sufficient anti-automation controls to detect and protect against data exfiltration, excessive business logic requests, or denial of service attacks. |  | ✓ | ✓ | 2.0 |
+| **15.7** | Verify the application has sufficient anti-automation controls to detect and protect against data exfiltration, excessive business logic requests, excessive file uploads or denial of service attacks. |  | ✓ | ✓ | 2.0 |
 
 ## References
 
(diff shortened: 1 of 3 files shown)


tghosth:
Here are responses to the points raised in #284 

> "Obtained from untrusted sources are stored outside the web root and limited permissions." -> "Untrusted file data obtained from untrusted sources are stored outside the web root and limited permissions"

Was already fixed.

> 16.3 Adding the need for AV I think is excessive as using a mime-aware library such as Apache Tika should largely remove the need for AV unless you allow users to upload dangerous file types such as EXEs or files that support active content (PDFs, Office documents, etc.)

I think that the likelihood of needing to allow PDF or office docs is high so I think this needs to stay.

> size limits + number of files uploaded limits
See: https://github.com/OWASP/ASVS/commit/88b77b5ed52bc04e6ae249771d707b9b3adf3cfb and https://github.com/OWASP/ASVS/pull/377/commits/6d3aea200d88ba818adf0d8777258775d3e1adb6

> quota limits per user (if needed)
> possibly using a different sub-domain to serve user uploads to prevent XSS
> the uploading of in-intentionally uploading executable code such as SVG images which are then served directly back to users

danielcuthbert:
This is solid, wow. we are damn close now, thank you

jmanico:
I think server side AV is a base requirement especially if a wide variety of files can be uploaded, like a dropbox.

PR #407, opened 2019-02-12
Proposed rewording of 2.1.8

m8urnett:
Verify the use of weak authenticators--such as SMS and email--is limited to secondary verification and transaction approval and not as a replacement to more secure authentication methods. If offering weak authenticators, verify that stronger methods are made available first, that users are aware of the risks, or that proper measures are in place to limit the risks of account compromise.


tghosth:
Thanks @m8urnett, I made some minor changes before merging so let me know if you think there were any problems with the final merge.

m8urnett:
I would suggest my original wording rather than changing it to the passive voice. It's generally better to avoid passive in technical writing.

tghosth:
@m8urnett is this better https://github.com/OWASP/ASVS/commit/4f4889f73848404bf24c904802c81746a56c4650

PR #409, opened 2019-02-12
Changes to ASVS 4.0

javixeneize:
Hi

I have added three new requirements, all of them only for L3:

1.14 Validate that there are security requirements that can be used as an input to test the undesired behaviour of the application.

2.5.6  Allow users to set a passphrase as their password

2.5.7 Recommend customers to use a password manager to store credentials securely


tghosth:
Hi @javixeneize, unfortunately all suggested changes need to be made into the markdown files in [this folder](https://github.com/OWASP/ASVS/tree/master/4.0/en). I am going to close this PR.

PR #410, opened 2019-02-13
Few minor readability tweaks in first two sections

grogsaxle:
elucidate?  Who uses that term?  

Hope the small changes are helpful.


jmanico:
There is nothing elucidate about elucidate per most folks vocabulary.

PR #437, opened 2019-02-15
Password length proposed change

tkisason:
I know that 5.1.1.2 requests at least 8 characters, and i know that by proper use of password hashing (scrypt, argon etc) with salting that the issue of short passwords can be of reduced risk, but people are people. Can we at least insist on something that's not breakable trivially? Today with modern GPUs anything with 8 characters can be bruteforced in under a day (https://twitter.com/hashcat/status/1095807014079512579) can we at least bump this up 50% to 12 characters, since that would drive complexity for an average 40 character base set (lowercase,number,space!#_) to 2560000 times more complexity than 8 characters. 16 would be ideal, but let's start somewhere. 

Sincerely,
Tonimir


tkisason:
Fixed merge conflict

PR #459, opened 2019-02-19
Fixed regex.

ronperris:
The new regex matches strings that contain 1 or more dots or numbers.

The json output appears to contain all the requirements now.


jmanico:
Thank you very kindly for this, Ron!

Issue #520, opened 2019-02-21
Appendix B references QA notes


vanderaj:
Add OWASP Top 10 2017,
Add OWASP Testing Guide 4.0
Add OWASP Proactive Controls (2018)
Add a link to the OWASP IOT project
Add a link to the OWASP Serverless Project

Add a link to the SecLists Github
Update PCI DSS to 3.2.1
Add PCI Secure Development Standard (2019)

Remove the reverse engineering guide (we don’t really refer to it in the ASVS). 

Move the Mobile links to their own section away from the main list as we’re referencing them rather than including them.

tkisason:
Done all of that and added the PCI Secure Software Lifecycle too, added OWASP Software Assurance Maturity Model too. 

See https://github.com/OWASP/ASVS/pull/533

vanderaj:
Thank you! :-) This is amazing work. :-) Please make sure you put yourself down in the Credits ticket so we can credit you for your work on this.

PR #533, opened 2019-02-23
Update 0x91-Appendix-B_References.md

tkisason:
Fixing issue https://github.com/OWASP/ASVS/issues/520


tkisason:
Ok, it seems i can't satisfy this linter no matter what i do. :/ The commit merges / renders perfectly though.

Issue #557, opened 2019-02-25
V2.1.7 - isn't it too aggressive for Level 1 to have this requirement?


pich4ya:
Refer to ASVS 4.0 draft version, V2.1.7 "Verify that passwords submitted during account registration, login, and password change are checked against a set of breached passwords either locally or via API, either using a zero knowledge proof or otherwise ensuring that the plain text password is not sent or used in verifying the breach status of the password. If the password is breached, the application must require the user to set a new non-breached password."

How many reliable free APIs for doing this job? I only know https://haveibeenpwned.com/API/v2. I don't think all apps need to have TB of breached passwords or use haveibeenpwned's API. Generally, it is overkill.

tkisason:
Suggested fix in: https://github.com/OWASP/ASVS/pull/558

tghosth:
Hi @pich4ya and @tkisason.

I believe this was made L1 because I marked it as being in the proactive controls but actually the proactive controls requirement doesn't require an external API so I have made this clear in the requirement here. Can you look at the wording change that I made here: https://github.com/OWASP/ASVS/commit/df20c0e6cf4debbceb668283199ea3584eefe34b and tell me if you think it is ok? I have closed #558 in the meantime.

PR #558, opened 2019-02-25

tkisason:
Fix for:
https://github.com/OWASP/ASVS/issues/557 

Merge to accept issue.


tghosth:
See my comment here: https://github.com/OWASP/ASVS/issues/557#issuecomment-467137750

tkisason:
I think this is reasonable... 👍

Issue #539, opened 2019-02-24
V6.3: Random GUID Generation


adamcaudill:
6.3.1 currently states the following (emphesis added):

> Verify that all random numbers, random file names, **random GUIDs**, and random strings are generated using the cryptographic module's approved random number generator when these random values are intended to be not guessable by an attacker.

GUIDs may be created using a variety of algorithms and PRNGs - if they need to be secure, they should only use the GUID v4 algorithm, and a CSPRNG. In many cases, developers are not aware of the differences, or what PRNG is being used to produce them. To ensure that this is being done properly, I would suggest the following addition:

> Verify that random GUIDs are created using the GUID v4 algorithm, and a cryptographically-secure pseudo-random number generator (CSPRNG). GUIDs created using other pseudo-random number generators may be predictable.

tghosth:
Created #565 rather than directly merging as I don't know if @vanderaj wants to accept new requirements at this stage.

PR #565, opened 2019-02-25
Add a requirement regarding random GUIDs as per #539

tghosth:


vanderaj:
I've applied this manually. There was a change to 6.3.1 that conflicted and it was too difficult to merge once that was in place.

Issue #535, opened 2019-02-23
V12.6.1 - open redirect description in SSRF subcategory


elarlang:
**subcategory**
V12.6 SSRF Protection Requirements

**requirement**
12.6.1 Verify that URL redirects and forwards only allow whitelisted destinations, or show a warning when redirecting to potentially untrusted content.

**problem**
Requirement 12.6.1 is not limited to Server-Side Request Forger and describes/covers Open Redirect (which is in most cases command to client)

**recommendation**
Fix the mismatch.
At the moment id does not feel correct but I don't have one-and-only clear recommendation how it should be fixed.
Requirement is not clear SSRF. If to watch it as open redirect (for browser), it does not even belong to V12 category.

pich4ya:
Agreed.

tghosth:
Opened #567 to address

tghosth:
This has now been merged

PR #567, opened 2019-02-25
Clarify SSRF and move unvalidated redirect to input validation

tghosth:
Remove unvalidated redirect from this section, clarify SSRF requirement and fix CWE


tghosth:
Based on #535

Issue #595, opened 2019-02-26
Appendix A - Transport Layer Security - not only for the Internet


securitybits:
> Transport Layer Security – Cryptographic protocols that provide communication security over the Internet

vs.

> Transport Layer Security – Cryptographic protocols that provide communication security over a network channel

tkisason:
Suggested fix in: https://github.com/OWASP/ASVS/pull/598

PR #598, opened 2019-02-26
Update 0x90-Appendix-A_Glossary.md

tkisason:
Suggested fix for https://github.com/OWASP/ASVS/issues/595


vanderaj:
Oops, I already applied this by hand. Must check pull requests before editing!

PR #623, opened 2019-03-06
updated for clarity of 0x20

grogsaxle:
12.1.1 dropped "attack".  probably just a personal preference

12.3.1 - Not entire sure what is trying to be conveyed here originally.  Tweaked some for what I think might be clarity, but I may be off base.


tghosth:
Merging into temp branch so I can propose edits :)

PR #629, opened 2019-03-26
Include possible API security mistakes

nmarher:
Include possible API Security mistakes


nmarher:
Hi.

We have seen that the tests give an error due to an alignment problem in the columns of the table. We have tried to fix it with remark-cli and we see that we get error in all the tables, not only in our changes.

It is possible that we are missing some kind of configuration, please check the content despite the problem of the table.

Thank you!

tghosth:
@nmarher there seems to have been a binary file added with this commit, please can you remove it from the commit. Also, are you sure that none of these controls appear elsewhere?

tghosth:
> Hi @nmarher, thanks for opening the PR. there seems to have been a binary file added with this commit, please can you remove it from the commit. Also, are you sure that none of these controls appear elsewhere?

Hi @nmarher, are you able to address his?

tghosth:
Gonna close this for now as no response from submitter

PR #650, opened 2019-05-19
resolve #625

### LINKED ISSUE #625: 2.5.1 unclear description for system generated initial secret
It's unclear what is meant by this description
> Verify that a system generated initial activation or recovery secret is not sent in clear text to the user

Could somebody add more details on this requirement ?

tghosth:


tghosth:
Thanks, it was a hangover from the original item, I have removed that element now, what do you think @vanderaj?

PR #654, opened 2019-06-10
Resolve issue #653

xxbaemaxx:


tghosth:
This is a new requirement so pushing to 4.1

PR #671, opened 2019-09-12
resolve #663

### LINKED ISSUE #663: Add a cert pinning requirement (L3)?
Does ASVS cover desktop apps? If so, should we recommend certificate pinning for level 3? I think this is something that the application and not the OS will have to handle.

However, I note that the introduction says:

> As of version 4.0, the ASVS will focus solely on being the leading web apps and service standard, covering traditional and modern application architecture, and agile security practices and DevSecOps culture.

tghosth:
Borrowed the wording from the MASVS :)


danielcuthbert:
A solid addition, thank you

PR #678, opened 2019-09-26
Add check on the correct usage of safe HTTP methods (issue #672)

VincentDS:


VincentDS:
As agreed in #672, a new requirement should be added to assure the proper use of safe and unsafe methods. This will have a reasonable impact on CSRF mitigations if browsers will set the SameSite flag automatically to Lax (If not specified otherwise).

PR #688, opened 2019-10-25
Persian Translation

SajjadPourali:
Add Persian Translation of OWASP Application Security Verification Standard 4.0.1


danielcuthbert:
I sadly can't read Persian but I do love how beautiful it looks. Thank you for taking the time to do this and doing a pull request.

PR #689, opened 2019-10-26
fix(script): sorted csv output + fixed linting

hashier:
Currently the csv generator script pics the files in random order and generates the csv file. This fix sorts the files before printing.

### DIFF
--- a/4.0/generate-csv.py
+++ b/4.0/generate-csv.py
@@ -5,6 +5,13 @@
 import re
 import os
 
+def atoi_if_possible(string):
+    return int(string) if string.isdigit() else string
+
+
+def string_num(string):
+    return [atoi_if_possible(el) for el in re.split(r'.*-V(\d+).*', string)]
+
 
 def area_from_filename(filename):
     filename_parts = filename.split("-")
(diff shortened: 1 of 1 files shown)


tghosth:
So the sorting thing is strange, when I run the script on Windows, the sorting seems fine. Is it possible this is some sort of Operating System setting. Either way, it doesn't seem to hurt so it is fine either way.

Please can you add a little more detail about the other changes you made? One was to add a header, what was the other? Thanks :)

hashier:
I guess it might be OS depending, the documentation is pretty clear that you can't rely on it though:


```
os.listdir(path='.')

Return a list containing the names of the entries in the directory given by path.
The list is in arbitrary order, [...]
```

2 main things.
- Sorting
- Fixing linting

The linting then again has 2 parts (:
- Adding a header (1st line of the document that names the fields)
https://github.com/OWASP/ASVS/pull/689/files#diff-1e224efd0940e696116bd0689eecdc2cR52

- Making sure that every line has the same amount of fields
https://github.com/OWASP/ASVS/pull/689/files#diff-1e224efd0940e696116bd0689eecdc2cR41-R42

tghosth:
Ok, great thanks!

PR #694, opened 2019-11-14
Update OWASP Application Security Verification Standard 4.0-en.csv

borislav-c:
Mapping of the requirements to OWASP Top 10 2017, using CWE classification: https://cwe.mitre.org/data/definitions/1026.html.


tghosth:
Hi @borislav-c. Thanks for the useful mapping. The CSV is automatically generated by a script so I don't think this is the ideal place for the mapping.

Could you maybe submit a separate file with the mapping and we can think about how we integrate the mapping afterwards?

borislav-c:
Hi @tghosth . Sure, NP.

borislav-c:
Hi @tghosth ,
Done (new file added and changes to the original reverted).

hashier:
@tghosth Talking about the CSV script; I created a PR (https://github.com/OWASP/ASVS/pull/689) to fix some problems with the CSV generation.

The current script writes out the CSV file in a random order + fixes a missing header in the CSV file.

hashier:
I suggest as well then to run the script instead of manually fixing the CSV file though.

tghosth:
Thanks, I want to have a look through the mapping before I merge, I will try and get to that as soon as I can but it may be a few weeks as it is busy season for consulting now :)

jmanico:
I'm going to merge but we need automation to fix this so its automatic in the future

PR #702, opened 2019-12-13
CORS = Cross-Origin ... and not Cross-Domain ...

cnotin:


tghosth:
Makes sense to me. Good spot :)

Issue #703, opened 2019-12-17
ASVS 4.0 - v14.5.1 Prefer "Handle all methods"


thatsjet:
I debated forking and submitting a PR but instead opted for some discussion first. If consensus is reached on a change I'm happy to draft it and submit.

My issue is with

`
14.5.1 | Verify that the application server only accepts the HTTP methods in use by the application or API, including pre-flight OPTIONS.
`

I appreciate the need for protection against verb tampering, but would prefer to see application creators consider attack driven logging in their design. In my company I teach developers to handle all valid request methods, but for methods that are invalid for the application context they should log the an invalid attempt while returning what would appear to be a valid response to the attacker. This both confuses the attacker, while providing high-fidelity logging and alerting, thereby providing an opportunity to respond to a bonafide attack.

Perhaps:
`
14.5.1 | Verify that the application server handles handles all HTTP methods including pre-flight OPTIONS and logs/alerts on on any requests that are not valid for the application context.
`

Thoughts?

tghosth:
What are your thoughts on the following?

`
14.5.1 | Verify that the application server only accepts the HTTP methods in use by the application/API, including pre-flight OPTIONS, unless it specifically logs/alerts on on any requests that are not valid for the application context.
`

jmanico:
Perhaps change..

unless it specifically logs/alerts on on any requests that are not valid for the application context

To

And log/alert on on any requests that are not valid for the current user

--
Jim Manico
@Manicode
Secure Coding Education
+1 (808) 652-3805

thatsjet:
@tghosth I think I like Jim's response better with "and log/alert." I think it's fine not to specify how to handle the response, but by expecting a log for invalid methods we automatically assume a handler.

So:
```
14.5.1 | Verify that the application server only accepts the HTTP methods in use by the application/API, including pre-flight OPTIONS, and log/alerts on any requests that are not valid for the application context.```

tghosth:
Ok great, can you open a PR?

thatsjet:
PR #705 opened

JasmineScholz:
How can we test this case?

elarlang:
Commenting closed issue may be not noticed.

"How to test", you should most likely search from other projects like testing guide and cheat sheet series:
https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods

PR #705, opened 2020-01-07
Resolve issue #703 - Log/alert on invalid methods.

thatsjet:
Resolves issue 703 to include language related to logging attempts to invoke an invalid method.


tghosth:
Great, thanks @thatsjet!

PR #727, opened 2020-04-03

csfreak92:
2.2.3 and 2.5.5 are talking about the same thing. Suggestion to remove 2.5.5 to simplify it since 2.2.3 covers the same requirement. Also, push notifications are only for mobile apps and it may not apply in this case because this is ASVS for web applications.


tghosth:
Thanks for your thoughts on this @csfreak92 

Would push notifications still be relevant where a mobile device is being used as a second factor for a web app? If you disagree please open an issue to discuss this.

Please open an issue with the text of both 2.2.3 and 2.5.5 so we can discuss if they are defintely duplicates.

PR #728, opened 2020-04-03
Update 0x12-V4-Access-Control.md

csfreak92:
Adding "insecure direct object reference" attacks in this control to point out that this could either be both authenticated and unauthenticated access instead of a generic attack.


tghosth:
Yeah, I think this makes sense. Thanks.

PR #731, opened 2020-04-12
Encourage SSRF prevention via libraries

devd:
SSRF is pretty hard to protect against by only validating inputs. For example, as currently phrased in the document, redirects might break the defense. Its much better to use SSRF safe libraries designed for protecting against SSRF than try to validate.

### DIFF
--- a/4.0/en/0x13-V5-Validation-Sanitization-Encoding.md
+++ b/4.0/en/0x13-V5-Validation-Sanitization-Encoding.md
@@ -33,7 +33,7 @@ Properly implemented input validation controls, using positive whitelisting and 
 | **5.2.3** | Verify that the application sanitizes user input before passing to mail systems to protect against SMTP or IMAP injection. | ✓ | ✓ | ✓ | 147 |
 | **5.2.4** | Verify that the application avoids the use of eval() or other dynamic code execution features. Where there is no alternative, any user input being included must be sanitized or sandboxed before being executed. | ✓ | ✓ | ✓ | 95 |
 | **5.2.5** | Verify that the application protects against template injection attacks by ensuring that any user input being included is sanitized or sandboxed. | ✓ | ✓ | ✓ | 94 |
-| **5.2.6** | Verify that the application protects against SSRF attacks, by validating or sanitizing untrusted data or HTTP file metadata, such as filenames and URL input fields, use whitelisting of protocols, domains, paths and ports. | ✓ | ✓ | ✓ | 918 |
+| **5.2.6** | Verify that the application protects against SSRF attacks, by using a library designed for requesting untrusted URIs with SSRF protections built in. | ✓ | ✓ | ✓ | 918 |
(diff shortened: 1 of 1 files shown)


devd:
Forgive me for directly going to a PR rather than discussing the change; the README suggested I should just create a PR so tha'ts what I did.

elarlang:
Please, no. It does not make sense for me.

How one library knows, what is good URL and what is not? You still need some whitelist and vaidation.

devd:
that's fair. Its hard to know which library to use. But, I feel like the rest of ASVS talk about sanitizing HTML and that has the same problem. My concern with the "check good URL proposal" is that redirects will trivially break it. You really have to introspect at a deeper level to be safe. See https://www.agwa.name/blog/post/preventing_server_side_request_forgery_in_golang

jmanico:
For SSRF defense I advise:

1) url validation for untrusted/user driven URL’s especially intra-server on the backend. But careful here the more complex rules you allow for url validation the more likely mistakes will be made. Keep it simple.
2) safe url construction, like url encoding fragments you add to intra server urls
3) note redirects AND forwards are potentially an issue here
4) lack of access control for back end API’s also weaponizes SSRF

While several libraries still have problems with URL validation problems encapsulating your SSRF defense into your framework is critical. If every dev does their own URL validation from scratch with unique regex per dev it’s unlikely to be a successful strategy, which I think is Dev’s point.

--
Jim Manico
@Manicode

jmanico:
Dev,

While compelling, I suggest this is not enough. What if the SSRF attack is against a feature url that is a valid feature and valid url but the user just doesn’t have access to it? Then access control is the underlying problem....

--
Jim Manico
@Manicode

devd:
Hmm..for those, the right fix to add to the blacklist via hooks provided by these libraries  For example, if I want to blacklist the IP per the case you pointed above . That IP needs to be added to the blacklist maintained by a library; we cant rely on URI validation because attacker.com DNS query can return an arbitrary IP or an arbitrary redirect (often handled without checking with the calling function for validation).

elarlang:
First, blacklist is bad idea. Blacklist or whitelist does not need to be in library, it can be in operation system level and in my opinion, it's better than library.

Library is just a tool and from that perspective this recommendation does not give anything extra. Tool need to be up to date and do not contain known vulnerabilities (kind of covers your idea) and you still need to know where and how to build your limitations to validate URL's or have some mitigation in place.

alex:
You cannot implement SSRF protections _on top_ of an existing HTTPS library, at least not easily. You need to be checking against the IP addresses that are having sockets opened to them. If you attempt to validate the URL before passing it to an HTTP library, things can be bypassed:

1. If you just look at the hostname, the attacker can use DNS to return an RFC 1918 IP for a public DNS name
2. If you do your own DNS resolution on the hostname, the DNS server can return a different response when the HTTP library does DNS resolution vs. when your validator does. (TOCTOU vulnerability)
3. If your HTTP library handles redirects internally, the server can just redirect to something with an RFC 1918 IP.

The comprehensive fix is to have an HTTP library that makes it easy to inject extra logic around sockets it opens.

https://www.agwa.name/blog/post/preventing_server_side_request_forgery_in_golang does a good job of walking through what this looks like in practice.

PR #738, opened 2020-04-17
Update 0x17-V9-Communications.md

csfreak92:
Compressing ASVS 4.0 - 9.1.1, 9.1.2, and 9.1.3 into one requirement since basically they are talking about the same thing. Verification for one applies for all as well.

### DIFF
--- a/4.0/en/0x17-V9-Communications.md
+++ b/4.0/en/0x17-V9-Communications.md
@@ -17,10 +17,8 @@ All client communications should only take place over encrypted communication pa
 
 | # | Description | L1 | L2 | L3 | CWE |
 | :---: | :--- | :---: | :---:| :---: | :---: |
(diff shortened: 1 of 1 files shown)


tghosth:
So I think there are edge cases when you would fulfil some but not all of these three requirements so I am going to close this suggestion for now. Happy to continue the discussion if you want to open an issue.

csfreak92:
Hi @tghosth, in which edge cases would one of them be fulfilled and not the others? I am not disagreeing with you on that, but I probably cannot see them right now. Maybe you can enlighten me with some of those cases?

tghosth:
You might have to support old ciphers even with new protocol versions or vice versa for legacy reasons.

You might support up-to-date ciphers and protocols when using TLS but be required for business reasons to also offer non-TLS connection

csfreak92:
But given that context, that application being assessed would be failing all three `9.1.1`, `9.1.2` and `9.1.3` ASVS requirements though. To me it still makes sense combining them though. Any thoughts for others to weigh in on this?

PR #742, opened 2020-04-23
Resolves #710

### LINKED ISSUE #710: 14.4.1 "every" content-type specifies a charset
This requirement should not be true for all content types.

The item 14.4.1 specifies that 
> Verify that every HTTP response contains a content type header specifying a safe character set (e.g., UTF-8, ISO 8859-1).

Per https://tools.ietf.org/html/rfc2046 this should only be for content types with a "text" subtype. For example, what `charset` should be used for `image/jpeg`?

tghosth:
This Pull Request relates to issue #710

### DIFF
--- a/4.0/en/0x22-V14-Config.md
+++ b/4.0/en/0x22-V14-Config.md
@@ -57,7 +57,7 @@ Configurations for production should be hardened to protect against common attac
 
 | # | Description | L1 | L2 | L3 | CWE |
 | --- | --- | --- | --- | -- | -- |
-| **14.4.1** | Verify that every HTTP response contains a content type header. text/*, */*+xml and application/xml content types should also specify a safe character set (e.g., UTF-8, ISO 8859-1). | ✓ | ✓ | ✓ | 173 |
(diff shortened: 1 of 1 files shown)


4.0/en/0x22-V14-Config.md
```diff
@@ -57,7 +57,7 @@ Configurations for production should be hardened to protect against common attac
 
 | # | Description | L1 | L2 | L3 | CWE |
 | --- | --- | --- | --- | -- | -- |
-| **14.4.1** | Verify that every HTTP response contains a content type header specifying a safe character set (e.g., UTF-8, ISO 8859-1). | ✓ | ✓ | ✓ | 173 |
+| **14.4.1** | Verify that every HTTP response contains a content type header. text/*, */*+xml and application/xml content types should also specify a safe character set (e.g., UTF-8, ISO 8859-1). | ✓ | ✓ | ✓ | 173 |
```
kingthorin, line 60:
```suggestion
| **14.4.1** | Verify that every HTTP response contains a content type header. text/*, */*+xml and application/xml content types should also specify a safe character set (e.g., UTF-8, ISO-8859-1). | ✓ | ✓ | ✓ | 173 |
```

PR #745, opened 2020-04-29
Resolves #741

### LINKED ISSUE #741: 14.5.3: Clarify not using rules matching URL prefixes or suffixes
I have seen time and time again mistakes implementing CORS origin whitelists around allowing access from subdomains. The current `14.5.3` states `uses a strict white-list of trusted domains` - I think this may cause confusion of organisations allowing access for all the subdomains of _trusted_ domains (including future subdomains), and even from other organisations subdomains. When implementing, they will use rules matching URL prefixes & suffixes, however often mistakes lead to access of malicious external domains. 

For example, an implementation allowing access to all domains ending in:

`this-site.com`

An attacker could register the domain for access:

`evilthis-site.com`

Or, an implementation granting access to all domains beginning with

`this-site.com`

An attacker could register:

`this-site.com.evil.com`

It's worth mentioning `uses a strict white-list of trusted domains, avoiding rules including prefixes or suffixes of trusted domains or subdomains`

i.e. `Verify that the cross-domain resource sharing (CORS) Access-Control-Allow Origin header uses a strict white-list of trusted domains, avoiding rules including prefixes or suffixes of trusted domains or subdomains and does not support the "null" origin.`

tghosth:
This Pull Request relates to issue #741


hazanasec:
Hi @tghosth, does the document not normally use the term 'subdomain' as opposed to 'sub-domain' ?

tghosth:
@hazanasec can you check and tell me?

hazanasec:
@tghosth ahh it seem's to use both sparingly, then I'm happy with the PR!

tghosth:
@hazanasec you are correct it should be 'subdomain'.

Issue #664, opened 2019-08-12
Responsible disclosure / Security.txt


commjoen:
Dear ASVS heroes,
as we are updating our project (MASVS) https://github.com/OWASP/owasp-masvs/issues/189, given existing documentation in OWASP that is no longer maintained, we stumbled upon the requirements regarding responsible disclosure. Now as I am sure that we should not focus on RD in the MASVS, we were wondering: is the ASVS the place to have requirements about it? 
With kind regards,
your mobile friends ^^.

danielcuthbert:
Hi

It's such a vast and conflicting subject, and I personally don't feel the ASVS needs to have it specifically listed. I love what this is doing https://www.thegfce.com/initiatives/r/responsible-disclosure-initiative-ethical-hacking/manifesto and would rather reference that.

davidclarke-au:
Please include https://securitytxt.org/ 

Full details - https://tools.ietf.org/html/draft-foudil-securitytxt-07

commjoen:
Thanks! We will pick this up in our issue https://github.com/OWASP/owasp-masvs/issues/283 :).

tghosth:
So my gut is that section 1 architecture is the best place

tghosth:
@davidclarke-au do you want to open a PR?

davidclarke-au:
> @davidclarke-au do you want to open a PR?

Will do, `1.1.8` seems the logical choice, so I'll aim for that.

vanderaj:
This is a new requirement so is a 4.1 merge

tghosth:
This got merged into master

elarlang:
Should we have markdown in requirement text? (Related issues #809, #810)

Need to consider, that those requirements are not present only in markdown / github and current solution is displayed:
```
Verify availability of a publicly available [security.txt](https://securitytxt.org/) file at the root or .well-known directory of the application that clearly defines a link or e-mail address for people to contact owners about security issues.
```

Proposal:
* no markdown in requirement text (keep only `security.txt`)
* helpful link (`https://securitytxt.org/`) move to references part

jmanico:
I strongly disagree with putting any markdown in requirements. A URL in the text is better, but markdown based links really harm the standard. And in general if a requirement needs a URL - it's likely not self explanatory. Can we please move ALL links to the bottom of the section in question so requirements are plaintext?

jmanico:
reopen per @elarlang

tghosth:
Fixed this with that commit

PR #752, opened 2020-05-04
Adding 1.1.8 - security.txt requirements

davidclarke-au:
Adding 1.1.8 - security.txt requirements to https://github.com/davidclarke-au/ASVS/edit/master/4.0/en/0x10-V1-Architecture.md

This Pull Request relates to issue #664


tghosth:
Made some minor tweaks, do you think they look ok?
Also, any idea if there is a relevant CWE number?

davidclarke-au:
> Thanks for the new PR!
> Made some minor tweaks, do you think they look ok?
> Also, any idea if there is a relevant CWE number?

LGTM, I think application is more appropriate. 

I could not find a relevant CWE for this, maybe https://cwe.mitre.org/data/definitions/1110.html thoughts?

tghosth:
How about:
https://cwe.mitre.org/data/definitions/1059.html

davidclarke-au:
That's the parent for 1110, so will probably work better, thanks.

davidclarke-au:
@tghosth I've added a commit to add CWE 1110 to 1.1.8

damienbod:
small note, the `security.txt` can also be placed at the root, not just the well known

davidclarke-au:
> small note, the `security.txt` can also be placed at the root, not just the well known

Quite correct @damienbod , I just used the terminology from the official site, I actually think the root is the better spot for it, as people may naturally search for `tld.com/security.txt`

I will add this reference too, thanks.

PR #754, opened 2020-05-04
Update 0x11-V2-Authentication.md - [ASVS 4.0] 2.9.2 control

csfreak92:
Removing L2 requirement for 2.9.2 as this is too much for it. It seems this is better to be for an L3 only requirement as it is kind of overkill for L2 since the assessor/tester/auditor would need verification details for manufacturer of HSM or TPM (cryptographic devices). If removing it from L2 is not agreeable, maybe we can move it to "Recommended but not required" o legend in L2 instead of ✓ "Required"?


tghosth:
Please open an issue for discussion

Issue #760, opened 2020-05-07
Add CWE hyperlinks to each CWE in the MD files


lfservin:
It would be a lot more useful to have hyperlinks to the cited CWE rather than have the reader do a search for each number found.

tghosth:
Did you write a script to do this? I think it is a good idea but it is a major change so we probably need to choose a sensible time just before a release to actually run it.

lfservin:
I used a sed one liner for this and edited all files with one command. Not too complicated. I didn't think it's such a big change, since the numbers stay the same. There's only a hyperlink to the corresponding cwe. The text displayed is the same.

lfservin:
the change was done at the md file-level. don't really know how it would look before a realease. The only sensible option would be to use pandoc to output a single md file out of all of them, run the script here, and then commit the change to the docx, pdf, etc. That wouldn't reflect on the json or probably the csv. 
If it's included now, then any new cwe added would need to follow the same format. The hard part is finding out the appropriate CWE. Making a link out of it is trivial in comparison.

tghosth:
I think it is more that it makes a lot of noise when diffing file versions as it changes every single line so I would prefer to do it as the last action before a release. Could you paste the command you used here?

lfservin:
this is the command:
`sed -Ei 's#(\| ?)([0-9]{1,4})( ?\|)#\1https://cwe.mitre.org/data/definitions/\2.html\3#' *.md`

I run it directly in 4.0/en/, but the path can be adjusted to run anywhere. 
it would be interesting to see in how much it is possible to run the same scripts as MASVS for building the documents in a docker container. ELse a MAkefile is the most straightforward way to go

tghosth:
great, thanks

tghosth:
I am going to close the PR for now and this issue as I have another reminder issue open.

elarlang:
I prefer to not have cwe's as links in md files.

https://github.com/OWASP/ASVS/issues/777#issuecomment-644989755

PR #761, opened 2020-05-07
Add links to the corresponding mitre CWE page for each CWE listed

lfservin:
This Pull Request relates to issue #760


tghosth:
I think I am going to close this. We need a different way to do this which doesn't put all the links into the original markdown. I think eventually we will have some sort of scripts to convert the markdown to other formats which will include coverting this sort of thing to links.

PR #771, opened 2020-05-18
Resolve #770

### LINKED ISSUE #770: Discussion of requirements 9.1.1-3 (from #738)
This issue continues a discussion from the closed PR #738 with @csfreak92 

[These requirements](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x17-V9-Communications.md#v91-client-communications-security-requirements) state:
> 9.1.1 | Verify that secured TLS is used for all client connectivity, and does not fall back to insecure or unencrypted protocols
> 9.1.2 | Verify using online or up to date TLS testing tools  that only strong algorithms, ciphers, and protocols are enabled, with the strongest algorithms and ciphers set as preferred.
> 9.1.3 | Verify that old versions of SSL and TLS protocols,  algorithms, ciphers, and configuration are disabled, such as SSLv2, SSLv3, or TLS 1.0 and TLS 1.1. The latest version of TLS should be the  preferred cipher suite.

I still believe these requirements should be considered separately but having read them more carefully, I think they also need to be clearer.

tghosth:
This Pull Request relates to issue #770

### DIFF
--- a/4.0/en/0x17-V9-Communications.md
+++ b/4.0/en/0x17-V9-Communications.md
@@ -17,9 +17,9 @@ All client communications should only take place over encrypted communication pa
 
 | # | Description | L1 | L2 | L3 | CWE |
 | :---: | :--- | :---: | :---:| :---: | :---: |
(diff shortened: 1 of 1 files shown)


tghosth:
Awaiting comments from @csfreak92

Issue #700, opened 2019-12-08
export.py script is broken for me


tghosth:
export.py script is broken for me and in the mean time we have "generate-json.py" and "generate-csv.py" which seem to work better.

How do we want to proceed?

lfservin:
would a "Makefile" be the way forward? it would be a more standard way to generate the target files. These "targets" could be created:
* pdf
* doc
* xls
* json
* csv
* all (all of the previous)

It would take all the guessing away. Could be matched with a "configure" file to make sure that python and pandoc are installed.

tghosth:
Sounds good in principle. I believe the [MSTG](https://github.com/OWASP/owasp-mstg) have a clever way of doing this as well? Did you look at that?

lfservin:
I just looked at that, but they seem to have a similar approach to the one in ASVS. they just have all across in a single directory called tools. It might be more convenient to have a makefile and the out the recipes for all other outputs there. 
Should I contribute one? I just need to understand which of all the scripts are still valid. There seems to be a bit of overlap between them.

tghosth:
"generate-json.py" and "generate-csv.py" seemed to work for me but export.py did not. You may need to experiement. Curious to see what you can come up with though :)

lfservin:
i'll create an issue. unfortunately the solution might involve moving the scripts into another directory to "unpollute" the root directory. This is against your pull request guideline. If I do it, I'd love to see it merged (at some point) to the directory, not just rejected. If you say it's ok, I'll create the issue and pull request

tghosth:
Which pull request guideline? I didn't understand what you mean.

lfservin:
this is the text of the pull_request_template.md:
`

This Pull Request relates to issue #...`
here's the file: https://github.com/OWASP/ASVS/blob/master/.github/pull_request_template.md

on a closer read, I might have mixed it with the MASVS guideline. I'll then make a pull request with a tools directory and a Makefile to be able to build this.

mgargiullo:
I ended up taking the export script a bit farther to use in one of our
reporting engines. I built an ASVS Python module that returns an object.

The object contains a list of Objectives as objects, each contain a list of
Sections as objects, which each contain a list of requirements objects.

From there, I can build any format I need.  Let me know if you'd like to
take a look at it... Maybe you can use it.

tghosth:
@mgargiullo this sounds like it would be useful. 
[...]

Issue #777, opened 2020-05-21
We should run the CWE script as one of the last actions


tghosth:
CWE script should hopefully be in issue #760

lfservin:
as already mentioned in #760 here's the script:
`sed -Ei 's#(\| ?)([0-9]{1,4})( ?\|)#\1https://cwe.mitre.org/data/definitions/\2.html\3#' *.md`
the execution path needs to be adjusted, since I ran it directly in 4.0/en

elarlang:
Do we actually need those links to *.md files (to data)? If links are required for pdf/docx/... output, generate it only for them. From my point of view it's another thing to clean up to use ASVS as data.

As previous discussion is in #760, but issue is closed, I link it here.

I have opened a bit related issues #809 and #810.

tghosth:
Personally I like the idea from a convenience point of view but I admit I don't know the impact on the scripts which convert to differenf formats. What specific conversion to data format are you thinking of?

elarlang:
The main question for all previously mentioned issue is - is this github repository more like some document or clear source of data. If it is document, then ok, we can have links. But all the tools which want to use this data later, need to parse text out from those links.

To create links to markdown here is the same like convert data to HTML specialcharacters before inserting data to database - it makes it convenient to display later back to HTML but it's done in wrong place - if you want to use this data somewhere else than HTML, it is in wrong format.

Like you can write one line to create this link to data, I can write one line to remove this link for me. Not a huge problem. But does it make sense?

tghosth:
I think the aim is that the markdown should look good and have links and stuff so that people can refer to it directly

lfservin:
The idea behind having the script run when "compiling" the documents is to not pollute the md files in the repository. So there should be no impact on tools parsing the existing md files.

tghosth:
so I kinda wanted them in the MD as well but maybe we need to pursue a gitbook or something as well instead an apply the links on "compilation". I'll think more about this.

lfservin:
You can look at my fork's branch for the status quo of the Makefile. I'm almost there. A word file can be generated, same for a pdf. it seems though as if the pdf is so far made by hand from the word file, since I couldn't fine where it is currently created. Here's the link to my WIP: https://github.com/lfservin/ASVS/blob/owasp-makefile/4.0/Makefile
it should be done before june 30th. 
other formats shouldn't be a problem

Not yet ready for a pull request, but maybe in a day or so

lfservin:
[...]

PR #815, opened 2020-06-26
Owasp makefile

### LINKED ISSUE #786: [Proposal] Avoid / minimize page breaks inside content tables in PDF version
I would propose to try to avoid/minimize page breaks inside content tables.

This would increase the overall page count but in my humble opinion also increase readability.
I don't think it's printed too often but read on the screen. Not being interrupted by footer/header would increase readability for me.

Example:

### LINKED ISSUE #787: [Proposal] Consider providing an ODF version
While Microsoft Word has without question an enormous user base, it might be a good thing for an open community to provide its documents in an open format.

If technically feasibly, it would be nice to also see an ODF version.

But this would really just be a nice to have.

lfservin:
This Pull Request relates to issue #700 , #777 
It might also help fix #786 
Does not yet fix #787


jmanico:
Can you take a look at this @tghosth when you get a chance?

lfservin:
I need to commit the makefile. It runs already on my machine. Sorry for the
delay.

jmanico:
Hey @vanderaj or @tghosth can you kindly look at this for us please?

tghosth:
@lfservin are you able to tidy this up and also respond to [my questions](https://github.com/OWASP/ASVS/issues/700#issuecomment-689131202)?

lfservin:
Yes, I am able to tidy it up.

The makefile runs in a build environment created from a Dockerfile. I
haven't created an official docker image out of it, but the effort of
creating it from a Dockerfile is minimal. It then runs inside it and tidies
up. Output are all the file formats.

I'll look later to see if i forgot a question to answer.

lfservin:
This pull request should definitely fix 
#700 : the new export.py is used within the makefile
#777: hyperlinks are added only to the files in the build directory, leaving the original ones in peace

We need to see the results, to determine whether this also takes care of #786 

Given an odf reference document, like the word document, an odt target is defined within the targets in the makefile. Thus there is very little left to be done for #787 

I've included a docker directory and a Dockerfile within it. This helps to have the "build environment" in a container, so that anyone can reproduce the results without breaking a sweat. The only requirement is to have docker installed. 

I also included a HowToBuid.md file to show the steps necessary to build the container and then the document. I would recommend to have an official hub.docker.com image under owasp, rather than having my own referenced there. That's the reason I'm only providing the dockerfile. 

There are two pdf versions. One uses "eisvogel.tex" for doing some latex magic to produce a nice-looking document.
The 0x00header.yaml and 0x01header.yaml files are the headers needed by pandoc to setup the document title and logo and all that. therefore the markdown file 0x00header.md is no longer necessary. Once a version of the pdf is chosen, we can remove one or the other. 

The makefile is outputing all document formats: pdf, csv, docx, flat.json, json, xml. 
As far as I can see the existing script is not creating the pdfs in other languages. This could be added to the Makefile if deemed necessary.

4.0/HowToBuild.md
```diff
@@ -0,0 +1,22 @@
+# How to create the document
+1. Install the Docker Daemon on your computer (Google is your friend)
+2. If running WSL or WSL2 make sure that you can talk to the Docker Daemon from the console (again, Google this)
+3. Run this command to create the container:
+
+```
+docker image build --tag documentBuilder -f docker/Dockerfile .
```
ike, line 7: Docker tags have two parts, a repository (with optional org), and then the tag, separated by a colon. For example: `asvs/documentbuilder:version1.0`. One of the restrictions is that the repository portion of the tag must be lowercase. More information is available [on the docker docs here](https://docs.docker.com/engine/reference/commandline/tag/#extended-description).

```suggestion
docker image build --tag asvs/documentbuilder:latest -f docker/Dockerfile .
```

4.0/HowToBuild.md
```diff
@@ -0,0 +1,22 @@
...
+```
+
+4. Run this command to run the container. The Volume you are mounting (`-v `) needs to be shared in the docker settings console for this to work:
+
+```
+docker run --rm -v "/Path/to/the/repo/4.0:/data" documentBuilder
```
ike, line 13:
```suggestion
docker run --rm -v "/Path/to/the/repo/4.0:/data" asvs/documentbuilder:latest
```

4.0/Makefile
```diff
@@ -0,0 +1,77 @@
+BUILDDIR=build
+TOOLSDIR=tools
+SOURCEDIR=en
+TEMPLATEDIR=templates
+SOURCES = $(SOURCEDIR)/*.md
+TARGETS=$(addprefix $(BUILDDIR)/,$(wildcard $(SOURCES)))
+TARGETNAME=OWASP Application Security Verification Standard 4.x
```
ike, line 7: Would the version number on this name be hand-updated for each release or could it be generated?

4.0/Makefile
```diff
@@ -0,0 +1,77 @@
...
+$(BUILDDIR)/en/%.md: $(BUILDDIR)
+	sed -E 's#(\| ?)([0-9]{1,4})( ?\|)#\1[\2](https://cwe.mitre.org/data/definitions/\2.html)\3#; s#^(\| :?---:? \| :?)---( .*)#\| :-----: \| :---------------------------------------------------\2#; s#.(./images/)#\1#' $(subst build,.,$@) > $@
+	 
+md: $(BUILDDIR)
+	pandoc $(PANDOC_MD_FLAGS) -o $(BUILDDIR)/source.md $(BUILDDIR)/$(SOURCES) 
+	sed -Ei 's#../images/#./images/#' $(BUILDDIR)/source.md
+
+pdf: 
+	pandoc $(PANDOC_PDF_FLAGS) --include-in-header=$(TEMPLATEDIR)/header.tex  -o "$(BUILDDIR)/$(TARGETNAME).pdf" $(BUILDDIR)/en/*.md $(SOURCEDIR)/0x00-Header.yaml
+	pandoc $(PANDOC_PDF_FLAGS) --include-in-header=$(TEMPLATEDIR)/header-eisvogel.tex -o "$(BUILDDIR)/$(TARGETNAME)2.pdf"  --template $(TEMPLATEDIR)/eisvogel.tex $(BUILDDIR)/en/*.md $(SOURCEDIR)/0x01-Header.yaml
```
ike, line 49: Will these two templates eventually converge? If not, we might want to consider a clearer filename for the second -- "2" -- file.

4.0/Makefile
```diff
@@ -0,0 +1,77 @@
...
+csv:
+	python3 $(EXPORT_TOOL) --format csv --language "$(SOURCEDIR)" > "$(BUILDDIR)/$(TARGETNAME).csv"
+
+xml:
+	python3 $(EXPORT_TOOL) --format xml --language "$(SOURCEDIR)" > "$(BUILDDIR)/$(TARGETNAME).xml"
+
+
+
+clean:
+	rm -rf $(BUILDDIR)/en
```
ike, line 76: The `en` directory still exists in my `build` directory, even after multiple runs of the script.

ike (commented):
Love this! The docker image built successfully on my machine, and the built files appear to be well-formed and complete. I left a few comments and suggestions below.

4.0/Makefile
lfservin: I would rather delete the one not using Eisvogel. No need to keep it. The decision has been taken

Issue #649, opened 2019-05-17
Web Sockets ?


c69:
There are no mentions of Web Sockets in current version of document (4.0)

As you can guess - very often Websockets are being a second-class citizen, missing some (or all) security mechanisms of their REST counterparts, like schema validation, 1-time use request tokens, data sanitation and so on.

I believe '13.5 Web-Sockets' section should be added to standard.

tghosth:
You make a good point. Could you open a PR with some specific controls?

vanderaj:
This would be a 4.1 merge, not a 4.0.2 merge

vanderaj:
Hi @c69 can you please let us know if you're in a position to create a PR in the proposed sections?

Racater:
Since there was no update on this issue for the last 3 months, I took the matter in hands and created a pull request. This is based on OWASP recommendations and multiple pentests done in our systems. Please suggest improvements or ask for clarifications.

jmanico:
We need the following changes:

1) 13.5.4, 13.5.5, 13.5.6 all need to start with "Verify that".
2) Whitelist needs to change to allow list
3) "you are using" looks like odd language.

I'm going to make these changes myself, I am on it.

jmanico:
Language fixed per dbfd721 and closing

PR #842, opened 2020-09-17
[v4.1] Added WebSocket subsection #649

Racater:
This Pull Request relates to issue #649

### DIFF
--- a/4.0/en/0x21-V13-API.md
+++ b/4.0/en/0x21-V13-API.md
@@ -60,13 +60,13 @@ Note: Due to issues with XXE attacks against DTDs, DTD validation should not be 
 
 | # | Description | L1 | L2 | L3 | CWE |
 | :---: | :--- | :---: | :---:| :---: | :---: |
-| **13.5.1** | Ensure that only authenticated users can open WebSocket connections. | ✓ | ✓ | ✓ | 306 |
-| **13.5.2** | Ensure that WebSocket Secure (wss) is used for all WebSocket connections. | ✓ | ✓ | ✓ | 319 |
-| **13.5.3** | verify that the Origin header is checked in the initial HTTP WebSocket handshake. | ✓ | ✓ | ✓ | 346 |
+| **13.5.1** | Verify that only authenticated users can open WebSocket connections. | ✓ | ✓ | ✓ | 306 |
+| **13.5.2** | Verify that WebSocket Secure (wss) is used for all WebSocket connections. | ✓ | ✓ | ✓ | 319 |
+| **13.5.3** | Verify that the Origin header is checked in the initial HTTP WebSocket handshake. | ✓ | ✓ | ✓ | 346 |
 | **13.5.4** | Verify that WebSocket messages are properly sanitized and encoded. | ✓ | ✓ | ✓ | 116 |
-| **13.5.5** | Ensure rate limiting is in place for WebSocket messages | ✓ | ✓ | ✓ | 770 |
-| **13.5.6** | If you are using session or channel tokens in webSockets, ensure that the tokens are obtained from a response to secure POST request only, and are not sent by the server through the Websocket connection. | ✓ | ✓ | ✓ | 319 |
-| **13.5.7** | If you are using session or channel tokens in webSockets, ensure that the tokens possess at least 64 bits of entropy and are generated using approved cryptographic algorithms. | ✓ | ✓ | ✓ | 331 |
+| **13.5.5** | Verify that rate limiting is in place for WebSocket messages | ✓ | ✓ | ✓ | 770 |
+| **13.5.6** | If you are using session or channel tokens in webSockets, verify that the tokens are obtained from a response to secure POST request only, and are not sent by the server through the Websocket connection. | ✓ | ✓ | ✓ | 319 |
+| **13.5.7** | If you are using session or channel tokens in webSockets, verify that the tokens possess at least 64 bits of entropy and are generated using approved cryptographic algorithms. | ✓ | ✓ | ✓ | 331 |
 
 ## References
 
--- a/4.0/en/0x21-V13-API.md
+++ b/4.0/en/0x21-V13-API.md
@@ -62,11 +62,10 @@ Note: Due to issues with XXE attacks against DTDs, DTD validation should not be 
 | :---: | :--- | :---: | :---:| :---: | :---: |
 | **13.5.1** | Verify that WebSocket Secure (wss) is used for all WebSocket connections. | ✓ | ✓ | ✓ | 319 |
 | **13.5.2** | Verify that, during the initial HTTP WebSocket handshake, the Origin header is checked against a whitelist of authorized origins. | ✓ | ✓ | ✓ | 346 |
-| **13.5.3** | Verify that websocket input from the client is sanitized server-side before being processed | ✓ | ✓ | ✓ | 116 |
-| **13.5.4** | Verify that rate limiting is in place for WebSocket messages | ✓ | ✓ | ✓ | 770 |
-| **13.5.5** | If you are using session or channel tokens in webSockets, verify that the tokens are initially obtained from a response to secure POST request only, and are not sent by the server through the Websocket connection. | ✓ | ✓ | ✓ | 319 |
-| **13.5.6** | If you are using session or channel tokens in webSockets, verify that the tokens possess at least 64 bits of entropy and are generated using approved cryptographic algorithms. | ✓ | ✓ | ✓ | 331 |
-| **13.5.7** | If only authenticated users should be able to use Websockets, verify that authentication is done before opening the WebSocket. | ✓ | ✓ | ✓ | 306 |
+| **13.5.3** | Verify that rate limiting is in place for WebSocket messages | ✓ | ✓ | ✓ | 770 |
+| **13.5.4** | If you are using session or channel tokens specific to WebSocket, verify that the tokens possess at least 64 bits of entropy and are generated using approved cryptographic algorithms. | ✓ | ✓ | ✓ | 331 |
+| **13.5.5** | If you are using session or channel tokens specific to WebSocket, verify that the tokens are initially obtained from a response to secure POST request only, and are not sent by the server through the WebSocket connection. | ✓ | ✓ | ✓ | 319 |
+| **13.5.6** | If only authenticated users should be able to use WebSockets, verify that authentication is done before opening the WebSocket connection. | ✓ | ✓ | ✓ | 306 |
 
 ## References
 
--- a/4.0/en/0x21-V13-API.md
+++ b/4.0/en/0x21-V13-API.md
@@ -60,13 +60,13 @@ Note: Due to issues with XXE attacks against DTDs, DTD validation should not be 
 
 | # | Description | L1 | L2 | L3 | CWE |
 | :---: | :--- | :---: | :---:| :---: | :---: |
-| **13.5.1** | Verify that only authenticated users can open WebSocket connections. | ✓ | ✓ | ✓ | 306 |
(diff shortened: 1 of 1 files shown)


Racater:
Please review the changes, As per the issue, this should be added only to 4.1, not to 4.0.2. Feel free to suggest improvements or ask for clarifications.

Sjord:
> Ensure that

Most rules start with "Verify that". Is that better?

> | **13.5.3** | verify that

"Verify" should be with a capital.

> Verify that WebSocket messages are properly sanitized and encoded.

What does this mean exactly? Who should do the sanitation? What values are dangerous in a websocket message?

> Ensure rate limiting is in place for WebSocket messages

Is this required? I don't think we currently require this for HTTP requests. Why would it be more important for websocket requests?

> If you are using session or channel tokens in webSockets, ensure that the tokens are obtained from a response to secure POST request only, and are not sent by the server through the Websocket connection.

What is this about? Why is sending tokens in a websocket insecure?

> If you are using session or channel tokens in webSockets, ensure that the tokens possess at least 64 bits of entropy and are generated using approved cryptographic algorithms.

Does this need a websocket-specific rule, or is 3.2.2 sufficient?

Racater:
@Sjord thanks a lot for the review.

> Most rules start with "Verify that". Is that better?
I have changed all instances of "ensure" to "verify" and corrected the capitalization for consistency with the rest of the document.

> Verify that WebSocket messages are properly sanitized and encoded
This rule comes directly from the OWASP website recommendations: https://owasp.org/www-project-web-security-testing-guide/v41/4-Web_Application_Security_Testing/11-Client_Side_Testing/10-Testing_WebSockets . As websocket messages are an unsafe data input, it should of course be sanitized. Specific sanitation steps are application specific. The data sanitation must be server side.

> Ensure rate limiting is in place for WebSocket messages
Websockets messages are often less well handled by reverse proxies, such as NGINX. Indeed the websocket data is in tcp stream, not in transaction. Websocket data can probably not be separated by packet or transaction and can not be filtered this way. This makes DoS easier on websocket. However, since there is no mention of rate limiting for http, it makes sense to remove this item.

> If you are using session or channel tokens in webSockets, ensure that the tokens are obtained from a response to secure POST request only, and are not sent by the server through the Websocket connection.
The problem with the server sending the session token via websocket is that it bypasses the CORS protection
offered by the browser. Moreover, since starting a session (getting a session token) is a state-changing operation on the server side, a POST request should be used. See http://www.adambarth.com/papers/2008/barth-jackson-mitchell-b.pdf .

> If you are using session or channel tokens in webSockets, ensure that the tokens possess at least 64 bits of entropy and are generated using approved cryptographic algorithms.

This has been added as some applications use a different tokens for webSockets and for http. It is good that users check websocket tokens entropy and not just 3.2.2.

Thanks again for the review. Like you suggested, maybe some elements could be removed if deemed too specific or redundant.

4.0/en/0x21-V13-API.md
```diff
@@ -56,6 +56,18 @@ Note: Due to issues with XXE attacks against DTDs, DTD validation should not be
 | **13.4.1** | Verify that a query allow list or a combination of depth limiting and amount limiting is used to prevent GraphQL or data layer expression Denial of Service (DoS) as a result of expensive, nested queries. For more advanced scenarios, query cost analysis should be used. |  | ✓ | ✓ | 770 |
 | **13.4.2** | Verify that GraphQL or other data layer authorization logic should be implemented at the business logic layer instead of the GraphQL layer. |  | ✓ | ✓ | 285 |
 
+## V13.5 WebSocket Security Requirements
+
+| # | Description | L1 | L2 | L3 | CWE |
+| :---: | :--- | :---: | :---:| :---: | :---: |
+| **13.5.1** | Verify that only authenticated users can open WebSocket connections. | ✓ | ✓ | ✓ | 306 |
```
tghosth, line 63: Will web sockets only ever be for authenticated uses? Could they not also be used for unauthenticated purposes as well?

4.0/en/0x21-V13-API.md
```diff
@@ -56,6 +56,18 @@ Note: Due to issues with XXE attacks against DTDs, DTD validation should not be
...
+| **13.5.2** | Verify that WebSocket Secure (wss) is used for all WebSocket connections. | ✓ | ✓ | ✓ | 319 |
+| **13.5.3** | Verify that the Origin header is checked in the initial HTTP WebSocket handshake. | ✓ | ✓ | ✓ | 346 |
```
tghosth, line 65: Checked for what?

4.0/en/0x21-V13-API.md
```diff
@@ -56,6 +56,18 @@ Note: Due to issues with XXE attacks against DTDs, DTD validation should not be
...
+| **13.5.4** | Verify that WebSocket messages are properly sanitized and encoded server-side. | ✓ | ✓ | ✓ | 116 |
```
tghosth, line 66: Sanitization and/or encoding should be performed before input is used in a different context, e.g. incorporated into a file path, used in a SQL query, written back to a web page's HTML, etc. It is not clear what the purpose is here.

Do we mean input validation to ensure that messages received conform to expected formats?

4.0/en/0x21-V13-API.md
```diff
@@ -56,6 +56,18 @@ Note: Due to issues with XXE attacks against DTDs, DTD validation should not be
...
+| **13.5.5** | Verify that rate limiting is in place for WebSocket messages | ✓ | ✓ | ✓ | 770 |
+| **13.5.6** | If you are using session or channel tokens in webSockets, verify that the tokens are obtained from a response to secure POST request only, and are not sent by the server through the Websocket connection. | ✓ | ✓ | ✓ | 319 |
+| **13.5.7** | If you are using session or channel tokens in webSockets, verify that the tokens possess at least 64 bits of entropy and are generated using approved cryptographic algorithms. | ✓ | ✓ | ✓ | 331 |
```
tghosth, line 69: Would it be better to reference the session token guidance contained elsewhere in this standard? Are these requirements different to other session token requirements which we already have?

4.0/en/0x21-V13-API.md
```diff
@@ -56,6 +56,18 @@ Note: Due to issues with XXE attacks against DTDs, DTD validation should not be
...
+## V13.5 WebSocket Security Requirements
+
+| # | Description | L1 | L2 | L3 | CWE |
+| :---: | :--- | :---: | :---:| :---: | :---: |
+| **13.5.1** | Verify that only authenticated users can open WebSocket connections. | ✓ | ✓ | ✓ | 306 |
+| **13.5.2** | Verify that WebSocket Secure (wss) is used for all WebSocket connections. | ✓ | ✓ | ✓ | 319 |
+| **13.5.3** | Verify that the Origin header is checked in the initial HTTP WebSocket handshake. | ✓ | ✓ | ✓ | 346 |
+| **13.5.4** | Verify that WebSocket messages are properly sanitized and encoded server-side. | ✓ | ✓ | ✓ | 116 |
+| **13.5.5** | Verify that rate limiting is in place for WebSocket messages | ✓ | ✓ | ✓ | 770 |
+| **13.5.6** | If you are using session or channel tokens in webSockets, verify that the tokens are obtained from a response to secure POST request only, and are not sent by the server through the Websocket connection. | ✓ | ✓ | ✓ | 319 |
```
tghosth, line 68: Are you talking about initially receiving the token here, i.e. how it is produced in the first place? Please can you clarify here.

tghosth (changes requested):
Thanks for your efforts on this. Please can you review and action my comments.

4.0/en/0x21-V13-API.md
Racater: You are right. This part is a repetition of section 3.2 of the standard. However, websocket session tokens are often different than cookie session tokens. They should be **checked separately**.

What do you think would be the best solution?

- Reference section 3.2 in the 13.5.7 requirement?
    - Such as done in rule 3.2.3. It makes the requirement explicit, but still creates a bit of repetition
- Replace 13.5.7 by a sentence in the 13.5 section text?
    - It removes the repetition, but creates a "hidden" requirement in the paragraph text. As such we are not sure that this specific token will be checked.

Racater:
thank you for your review @tghosth . I will work on it and commit updates

4.0/en/0x21-V13-API.md
Racater: clarified

4.0/en/0x21-V13-API.md
Racater: Good point. Even though [some resources](https://www.freecodecamp.org/news/how-to-secure-your-websocket-connections-d0be0996c556/#-4-authenticate-users-before-ws-connection-establishes) say that "establishing a websocket connection is a bit expensive anyway", ASVS should in my opinion also allow unauthenticated websockets. I modified the rule.

Note that as stated in [this paper](https://www.researchgate.net/publication/315858773_Slowloris_DoS_Countermeasure_over_WebSocket), websocket is vulnerable to slowloris (not closing connections). It might be worth it to add another rule for connection limiting, to investigate...

4.0/en/0x21-V13-API.md
Racater: You are right. What do you think of such a formulation:

> Verify that websocket input from the client is sanitized server-side before being processed

This item is in part a repetition of section 5 of ASVS. The aim is to make sure that Websocket inputs are taken into account, as they are not mentioned in section 5. Maybe it's better to modify section 5 though?

4.0/en/0x21-V13-API.md
Racater: clarified

elarlang:
> V13.5.4| Verify that WebSocket messages are properly sanitized and encoded server-side.

Isn't it covered by 5.2.1, 5.3.1, 5.3.3? If not, what is the difference?

> V13.5.5 Verify that rate limiting is in place for WebSocket messages

Isn't it covered by 11.1.4? If not, what is the difference?

Racater:
@tghosth I updated the PR based on your comments. I also reordered the rules in a more logical order. 

Note that there is still repetitions as follows:
- Rule 13.5.3 is basically section 5 (Validation) applied to Websockets
- Rule 13.5.6 is rules from 3.2 (tokens) applied to Websockets

I am still a bit unsure how to handle the repetition and any suggestion would be welcome.

elarlang:
There are not clear rules (based on my knowledge) and there are some duplicate requirements (in files section for example) because of that.

I watch WebSocket as communication channel - it does not have anything to do how data is used later, I think we don't need separate requirement for "data from websockets must be validated / sanitized / converted also". It must be done anyway.

With anti-automation requirement it's not that clear, but in my opinion proposed requirement should be part of current 11.1.4.

> Rule 13.5.6 is rules from 3.2 (tokens) applied to Websockets

What is 3.2 here? Current subcategory 3.2 is "V3.2 Session Binding Requirements" and it does not make sense in this context.

Racater:
Good point, I removed the sanitation rule and added more details about WebSocket-specific tokens.

4.0/en/0x21-V13-API.md
Racater: I removed the rule altogether, to avoid duplication.

Racater:
done, waiting for your input

jmanico:
Is this ready to be merged?

elarlang:
Some feedback/comments on style/language:
* "whitelist" may be alarming nowadays, something like "allow list" or "allowed domains" instead
* Probably every requirement should start with "Verify that" like all other requirements in ASVS
* "you are using" -> "being used", "is in use"

jmanico:
I am fixing the language now - standby.

jmanico:
Language fixed per dbfd721

PR #846, opened 2020-09-30
Change related to #843

tghosth:
Changes made by @bretik but originally targeted to 4.0.2 instead of 4.1.

I want to finish the discussion on #843 before this gets actually merged.


jmanico (changes requested):
I believe 3.2.3 needs to be removed and this PR should be rejected as is. There is no defense from XSS no matter how you store tokens.

jmanico:
This is resolved elsewhere and we are also updating other requirements around storage of sensitive data so we are consistent. Closing for now.

ff49b9e

Issue #764, opened 2020-05-10
[Clarification/For Discussion] Default credentials mean in 2.10.2


csfreak92:
ASVS 4.0 - 2.10.2 verification requirement is:

`Verify that if passwords are required, the credentials are not a default account.`

I would like to clarify from the community/maintainers of this project regarding the term `default credentials` mean on this control? Does `default credentials` stand for some applications that have default credentials after installation? The way I interpreted this is if an application is deployed/installed and there are default credentials it fails this verification requirement (e.g. phpmyadmin application has "root", "admin" or something similar default credentials that comes with the installation of the application).

If my assumption and thinking is correct, I'll create a Pull Request to link this issue, else please help me understand this control better. This confuses our team a lot of times.

tghosth:
![](https://media.giphy.com/media/UVdiMrmmwTWDcN2dZ5/giphy.gif)

So I am not sure what this means and I am also struggling to trace it back to the relevant NIST requirement. 

@vanderaj it looks like requirement [2.10.2](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x11-V2-Authentication.md#v210-service-authentication-requirements) started life as part of 2.21 added in this commit: https://github.com/OWASP/ASVS/commit/e0aa0e60ee1a0a80fc762e1e5b377d6c6244489d

Any idea if that NIST link to 5.1.1.1 is accurate?

csfreak92:
Hi @vanderaj, @tghosth, any updates on this issue? Is there anything needed from my end?

danielcuthbert:
My bad, this was assigned to me and I've stalled. Originally it was just as you said, for example, admin/admin123, and in hindsight, I can now see how this is actually not helpful in the way it was written. I'd really like to see how you'd make it better, so yeah please submit a PR

csfreak92:
If there are no more ideas, then alright will submit a PR later today. I think there's a way we could make this better.

csfreak92:
@danielcuthbert, I analyzed it more about this requirement. It looks like this is about service authentication, though this ASVS requirement is still vague so I created a PR to make it better. Let me know your thoughts on it.

danielcuthbert:
thank you @csfreak92 ill check it this weekend!

tghosth:
Resolved by #848 and #859

PR #848, opened 2020-10-09
Update 0x11-V2-Authenticaftion.md

csfreak92:
Suggestion to make 2.10.2 better. 

This Pull Request relates to issue #764


Racater:
The credentials should avoid default values regardless of the account being service or not. I would remove the "service authentication" precision.

Sjord:
> Verify that if passwords are required for service authentication, the service account used is not running on default credentials. (e.g. root/root or admin/admin are default in some services during installation).

I think "running on" is not the proper verb here. Is it sufficiently clear what "default credentials" are? Should we say something like "Verify that passwords are changed from their default after initial installation."

In [another project](http://www.sjoerdlangkemper.nl/2020/09/30/iot-security-regulation/), I phrased it like this, to point more to the actual problem of default passwords:

> After initial setup, passwords must be unique for each device, or defined by the user.

tghosth:
I think it makes sense to focus specifically on service account as that is the point of this section.
@csfreak92 do you agree with the small change I made in 46fd8a0?

Racater:
> I think it makes sense to focus specifically on service account as that is the point of this section.
@tghosth That is a good point.

It is clear that default passwords should not be used for user accounts as well. If this rule is modified to specify "service accounts", a new rule should be added that mentions default passwords (and server-generated passwords). Otherwise, the standard will be weakened.

danielcuthbert:
All good my side and @tghosth is going to merge so as to not break things. Thanks for the effort @csfreak92 :)

PR #907, opened 2021-02-24
Building ASVS with Honkit

joubin:
As of late - we have seen lots of issues using Pandoc to do our builds. There are currently several manual steps that have to take place to get Pandoc to work to generated a PDF. 

I this merge request a few things are happening! 

- Restructure the format of the repo putting all older version of ASVS in `archive`
  - I haven't been involved in the reasoning to keep it? Maybe to have previous versions available for others? Maybe we move this to releases
- Use Honkit as a means for displaying and building the document 
- Scripts for generating PDF and EPUB
- Scripts for utilizing docker for the building process

Honkit is a fork of Gitbooks used for building web based books. It simplifies a few different things. 

To test the working setup:

You can simply run 

```bash
./honkit-serve.sh en
```

which will show you the following when you visit `http://localhost:4000`:

https://user-images.githubusercontent.com/1058767/108953847-b6a17e00-7620-11eb-851f-8aef512e435e.mov

There are also scripts that make it easier to generate PDF, EPUB, and MOBI formats of the content. 

For example, to generate a `mobi` file that can be added to a Kindle, we can run: 

```bash 
./honkit_generate_document.sh en mobi test.mobi
```

For `PDF`, we can run: 

```bash
./honkit_generate_document.sh en pdf test.pdf
```

Finally, I've also created a dockerized export for the `python` scripts. Many organizations have an approval process for adding decencies like `dicttoxml`. From a security perspective, this doesn't really change anything. However, from a compliance perspective, this may enable different workflows.


tghosth:
Hi @joubin, this is a massive massive change that I don't want to drop in all in one go.

Regarding the archive, it's is something we can think about although I am not sure o want to move anything in the 4.0 folder if it will mess up the file history.

Regarding output generation, we currently have a paid contractor working on creating an improved and unified build process for ASVS, MASVS and MSTG so I want to see where we are before we make any other modifications. On the other hand, I am definitely interested in a way of creating a gitbook style output format.

Can you create an issue for each separate aspect you discuss in this PR so we can see where we go with each?

joubin:
Absolutely @tghosth.

Issue #739, opened 2020-04-18
V2.5.7 - should be applied to L1


fujiokayu:
>V2.5.7 - Verify that if OTP or multi-factor authentication factors are lost, that evidence of identity proofing is performed at the same level as during enrollment.

If the application implements OTP or multi-factor authentication, the requirements of V2.5.7 is really important.
And V2.5.7 can be verified with DAST.

I think this item should be verified in L1.

bedirhan:
Same with 3.3.3

tghosth:
For reference, 3.3.3 is:

> Verify that the application terminates all other active sessions after a successful password change, and that this is effective across the application, federated login (if present), and any relying parties.

tghosth:
@jmanico what do you think about the claim that these should be L1 controls?

vanderaj:
This would represent a new requirement level change, and thus is a breaking change. 4.1

jmanico:
I think moving this to L1 is fundamental. I agree. Can we see PR's on this please? cc @fujiokayu  and @bedirhan  ?

fujiokayu:
@jmanico 
Thank you for checking this issue. I will send PR later.

fujiokayu:
Now that my PR has been merged, close this issue.
Thanks @jmanico , @bedirhan

PR #915, opened 2021-03-13
Include two existing requirements in L1

fujiokayu:
Applied 2 requirements to L1
- V2.5.7
  - Verify that if OTP or multi-factor authentication factors are lost, that evidence of identity proofing is performed at the same level as during enrollment.
- V3.3.3
  - Verify that the application gives the option to terminate all other active sessions after a successful password change (including change via password reset/recovery), and that this is effective across the application, federated login (if present), and any relying parties.
  
This Pull Request relates to issue #739


jmanico:
I appreciate your very precise language

Issue #889, opened 2020-12-28
Protect against HTTP request smuggling


Sjord:
If the proxy and application server keep an open connection and don't agree on where HTTP messages start and end, it may be possible to inject into the next HTTP request, a vulnerability known as [HTTP request smuggling](https://portswigger.net/web-security/request-smuggling).

This can be verified hands-on by trying several HTTP requests. It may be a little bit out of scope, since it concerns the infrastructure more than the application itself.

jmanico:
This is interesting research. Can you take this information and convert it into a set of remediation requirements? This paper provides a set of defensive instructions at the end that should guide how we write this up as requirements...

--
Jim Manico
@Manicode

jmanico:
Would you care to take the advice at the bottom of this article and provide one PR for this issue? cc @Sjord

Sjord:
Perhaps for the ASVS something like this is sufficient.

> Verify that the infrastructure handles each request separately, even with conflicting `Transfer-Encoding` and `Content-Length` headers, to prevent HTTP request smuggling.

This does not offer very concrete steps to resolve it. One way to solve request smuggling is to use HTTP/2 for the backend connection, because HTTP/2 has better separation of requests. However, pentesters can't really verify how the backend and the proxy are connected, or even if there is a proxy.

jmanico:
Because developers cannot take much action and testers can not easily discover this, I am inclined to let it go and drop the issue.

https://www.youtube.com/watch?v=L0MK7qz13bU

Sjord:
> Because developers cannot take much action

The application developers probably can't solve this issue. Someone from infrastructure or operations would need to fix the configuration of the proxy. However, this is also the case for TLS versions as in [9.1.3](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x17-V9-Communications.md#user-content-v91-client-communications-security-requirements:~:text=9.1.3,-%5BMODIFIED%5D).

> and testers can not easily discover this

It's not particularly hard to find. I have found it several times, and I think Burp detects it with its built-in scanner. However, the reverse is not true; it is pretty hard to verify that an application isn't vulnerable. But again, that is also true for XSS and many other vulnerabilities.

> I am inclined to let it go

I agree that this vulnerability is a bit unconventional. However, it can have a high impact. In some cases, an unauthenticated user can perform arbitrary authenticated requests, or steal the session token of another user.

jmanico:
In that case I like your requirement, let's PR this. :)

[...]

PR #931, opened 2021-03-19
Add request smuggling verification

Sjord:
Related to issue #889. Put in V14.5, since this vulnerability arises from a
conflict between the Transfer-Encoding and Content-Length request headers. Put
as L1 because of its possibly high impact, and it's possible to pentest it
remotely.


jmanico (approved):
I am going to move this to L2 after the merge

Issue #726, opened 2020-03-12
move V12.3.4 to V12.5 subcategory and rephrase?


Bankde:
Category / Subcategory
V12.3 File execution Requirements
V12.5 File Download Requirements

Requirement
`
12.3.4 Verify that the application protects against reflective file download (RFD) by validating or ignoring user-submitted filenames in a JSON, JSONP, or URL parameter, the response Content-Type header should be set to text/plain, and the Content-Disposition header should have a fixed filename.
`

Recommendation
I have read about RFD [here](https://www.trustwave.com/en-us/resources/blogs/spiderlabs-blog/reflected-file-download-a-new-web-attack-vector/) and [here](https://medium.com/@Johne_Jacob/rfd-reflected-file-download-what-how-6d0e6fdbe331). It doesn't look like file execution issue. In JSON exploit case, there isn't even a file at all. It think it's better to move this recommendation to V12.5 which focuses more on downloading file.

Also consider this situation:
1. Uploads malicious malware but names the file as "example.jpg".
2. Upload "example.jpg" into the server.
3. Trick other users to download that "example.jpg" as "example.exe".

I think it's very familiar to the requirement about Content-Type and Content-Disposition header. The current recommendation is a bit too focus on RFD attack so I think we could rephrase to make it more general to cover other cases.
`
Verify that the application validates or ignoring user-submitted filenames, including in a JSON, JSONP, or URL parameter. The response Content-Type header and Content-Disposition header should be fixed to the file and securely handled by the application instead of the user.
`

jmanico:
I suggest avoiding using a user submitted file name (as part of file upload) for storage. I suggest storing the uploaded file using the primary key of the table insert or similar (ir: trusted data) and using the user submitted filename for display purposes only. 

--
Jim Manico
@Manicode
Secure Coding Education
+1 (808) 652-3805

Bankde:
Hi. I know that's the file name handling best practice and is covered by V12.3.1. I think the current V12.3.4 requirement focuses on verifying that Content-Type and Content-Disposition are not abused or controlled by attacker during download process? That's why I suggest moving to V12.5 File Download Requirements subcategory.

jamesly123:
I think the fact that RFD allows execution of files on a victim's machine makes it appropriate for this test objective to remain under 'File Execution Requirements'.

Bankde:
[...]

Issue #721, opened 2020-02-22
Discussion: V14.4.2 - move or change - serving content as attachment


elarlang:
> V14.4.2 Verify that all API responses contain Content-Disposition: attachment; filename="api.json" (or other appropriate filename for the content type).

As it is API specific, move this requirement to V13.2

(goal ASVS v4.1)

jmanico:
Agreed ; good call.

--
Jim Manico
@Manicode

tghosth:
Marking for 4.1

jmanico:
We're ready for a PR on this @elarlang

elarlang:
With current wording it makes sense to move it, but maybe we should review and generalize the wording instead.

I don't know how to wordsmith it - idea is, if some content/file is not meant to be displayed and executed in browser as part of used application functionality, it must be served as attachment. Also related to requirement:
> V12.5.2 Verify that direct requests to uploaded files will never be executed as HTML/JavaScript content.

Far away from requirement proposal, open for discussion.

jmanico:
My instinct says something closer to:

V12.5.2 Verify that uploaded files cannot execute active web content such as HTML/JavaScript content.

Maybe? Bah this is tough.

Bankde:
I agree it looks more fit to 13.2 than 14.4

@elarlang There are some similarities to V12.5.2 but I think they are different.
V12.5.2 => Prevent browser to interpret file as Javascript/HTML (prevent XSS, etc.)
V14.4.2 => Prevent browser to download file as malicious file name; e.g. download user.json into user.bat. (prevent social engineering).
The file could include any script but is safe in its content-type/context and could be malicious in different context when browsers choose incorrect interpretation.

However, I'm more concerned about the reason of this requirement.
I don't think I have seen this recommendation (adding content-disposition header into every API responses) apart from [RFD ](https://medium.com/@Johne_Jacob/rfd-reflected-file-download-what-how-6d0e6fdbe331) mitigation, and I don't think I have seen top webs following this practice either?

Are there other reasons I'm missing or threats that this requirement mitigate?
If it's solely for RFD, and not a common recommendation/practice, I think L1, L2 sounds too stressed.

elarlang:
If you force browser to send content as attachment on direct access, then the content-type can be text/html and it will be not executed in browser. From that perspective 14.4.2 and 12.5.2 serve the same goal as well.

Bankde:
[...]

PR #932, opened 2021-03-19
issue #726

elarlang:
This Pull Request relates to issue #726


Bankde:
Thank you so much.

By the way, since the original requirement focuses on RFD and #721 is likely related to RFD too, there might be another wording change. It would be nice to wait the merge until the discussion on both issues be clear.

jmanico (approved):

Issue #885, opened 2020-12-18
Secure the postMessage interface


Sjord:
Websites can communicate to other websites loaded in other windows or iframes using the [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) interface. This presents an attack surface, which can lead to XSS. I think the ASVS should have one or more requirements that verify whether the postMessage interface is secure. A good start would be the [recommendations on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#Security_concerns).

jmanico:
+1

Esentially users of web messaging need to do at least 2 things.

1) Verify the origin of incoming messages
2) Treat incoming messages as untrusted data (a) verify their format (b) render those messages safely (like avoid innerHTML unless you DomPurify, use innerText, and all the other XSS defense advice) 

--
Jim Manico
@Manicode

jmanico:
hey @Sjord can we get a PR on this one from you?

Sjord:
I propose this:

> Verify that messages received by the `postMessage` interface are discarded if the origin of the message is not trusted, or if the syntax of the message is invalid.

DOM XSS through innerHTML is not particularly specific to messages, perhaps this already goes under [5.3.6](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x13-V5-Validation-Sanitization-Encoding.md#user-content-v53-output-encoding-and-injection-prevention-requirements:~:text=5.3.6). I think if the origin of the message is verified, it can be somewhat considered trusted data.

jmanico:
This is perfect. Let's PR this! please add [ADDED] to the text of the new requirement please! :)

https://github.com/OWASP/ASVS/blob/master/CONTRIBUTING.md

jmanico:
And if the label is confusing just ignore it and I'll take a standard PR and fix the label myself, whatever is easier for you.

elarlang:
Formating note: please don't use markdown syntax in requirement text (`postMessage`)

I'm not too familiar with postMessage, can you please explain me the meaning of:
* the origin of the message is not trusted
* if the syntax of the message is invalid

jmanico:
This is a well establish standard from the HTML5 era, it's old and well established.
See:  https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage
And: https://caniuse.com/?search=postMessage

Can you give this a quick read first and ask questions if it's confusing?

elarlang:
[...]

PR #937, opened 2021-03-22
Add requirement for verifying postMessage messages

Sjord:
Check origin and syntax. Marked this as CWE-346: Origin Validation Error.
CWE-20: Improper Input Validation would also apply.

This section, V4.2 Operation Level Access Control, is not the best fit for this
requirement, but we plan to move it to a better section later.

Related to issue #885.


jmanico (approved):

jmanico:
Thank you for your contribution!

Issue #952, opened 2021-03-30
Typo in Session Management 3.2.4


avasiljeva:
There is a typo in 3.2.4 of Session Management:
> Verify that session token are generated using approved cryptographic algorithms.

PR #953, opened 2021-03-30
Fix typo in Session Management 3.2.4

avasiljeva:
This Pull Request relates to issue #952.


jmanico:
Thank you for this! :)

avasiljeva:
Didn't get it clear if makes sense to pull it also into 4.0.3 branch, so change does not get lost in case it is released.
Could you please advise if it's enough with master or better to put it also into 4.0.3?

jmanico:
Just main for now, we'll take care of the 4.0.3 work! Thank you!

PR #973, opened 2021-04-19
Update 0x21-V13-API.md

jmanico:
make token size inline with other session standards


Sjord:
Should this requirement get the `[MODIFIED]` tag?

elarlang:
> Should this requirement get the `[MODIFIED]` tag?

No, tags are "diff from v4.0.2". Compared to v4.0.2 this requirement is still `[ADDED]` and it does not matter how many changes it will get.

PR #980, opened 2021-05-01
Deprecate TLS v1.2

### LINKED ISSUE #979: Deprecate TLS v1.2 for PCI-DSS v3.2.1 Requirement 2.2.3
I would like to propose that OWASP recommend the migration from TLS v1.2 to TLS v1.3 [or later] to support PCI-DSS v3.2.1 Requirement 2.2.3 reproduced below:

image

Therefore, [V9.1 Client Communications Security ](https://github.com/cmlh/ASVS/blob/master/4.0/en/0x17-V9-Communications.md#v91-client-communications-security) will need to be modified.

The alignment with PCI-DSS v3.2.1 is tracked by GitHub Issue https://github.com/OWASP/ASVS/issues/317#issuecomment-829077114

cmlh:
Fix #979

### DIFF
--- a/4.0/en/0x17-V9-Communications.md
+++ b/4.0/en/0x17-V9-Communications.md
@@ -13,13 +13,13 @@ Leading industry advice on secure TLS configuration changes frequently, often du
 
 ## V9.1 Client Communications Security
 
-All client communications should only take place over encrypted communication paths. In particular, the use of TLS 1.2 or later is essentially all but required by modern browsers and search engines. Configuration should be regularly reviewed using online tools to ensure that the latest leading practices are in place.
+All client communications should only take place over encrypted communication paths. Configuration should be regularly reviewed using online tools to ensure that the latest leading practices are in place.
 
 | # | Description | L1 | L2 | L3 | CWE |
 | :---: | :--- | :---: | :---:| :---: | :---: |
 | **9.1.1** | [MODIFIED] Verify that TLS is used for all client connectivity, and does not fall back to insecure or unencrypted communications. ([C8](https://owasp.org/www-project-proactive-controls/#div-numbering)) | ✓ | ✓ | ✓ | 319 |
 | **9.1.2** | [MODIFIED] Verify using online or up to date TLS testing tools that only strong cipher suites are enabled, with the strongest cipher suites set as preferred. | ✓ | ✓ | ✓ | 326 |
-| **9.1.3** | [MODIFIED] Verify that only the latest recommended versions of the TLS protocol are enabled, such as TLS 1.2 and TLS 1.3. The latest version of the TLS protocol should be the preferred option. | ✓ | ✓ | ✓ | 326 |
+| **9.1.3** | [MODIFIED] Verify that only TLS 1.3 is enabled. | ✓ | ✓ | ✓ | 326 |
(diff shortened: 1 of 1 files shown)


jmanico:
I see no reason to deprecate TLS 1.2 (yet) in ASVS. If there is a PCI rule or similar please provide a citation. As of today, ssllabs.com and other engines still allow TLS 1.2 at the A+ level and I see no standard or RFC deprecating TLS 1.2.

cmlh:
The [Australian Cyber Security Centre
](https://www.cyber.gov.au/acsc/view-all-content/publications/implementing-certificates-tls-https-and-opportunistic-tls) states "_TLS 1.3 removed vulnerable cipher suites found in TLS 1.2, while introducing stronger cipher suites._" @jmanico

jmanico:
They also say, “If organisations need to use older versions of TLS, care should be taken to select cipher suites and configure protocol features in accordance with this guide and the ISM.”

So no. We are not going to mandate TLS 1.3… yet.

cmlh:
> They also say, “If organisations need to use older versions of TLS, care should be taken to select cipher suites and configure protocol features in accordance with this guide and the ISM.” So no. We are not going to mandate TLS 1.3… yet.

Giving developers a "_preferred option_" will always be "_... vulnerable cipher suites found in TLS 1.2,_" as there is no urgency to upgrade without an enforced requirement @jmanico

jmanico:
While I get your point and think it’s fair - my vote on a TLS 1.3 only requirement today would be no. I’ll certainly vote yes to this in the future, but not now.

cmlh:
> While I get your point and think it’s fair - my vote on a TLS 1.3 only requirement today would be no. I’ll certainly vote yes to this in the future, but not now.

We can meet in the middle if we mandate this for web servers, languages and [web browsers](https://caniuse.com/tls1-3) that support TLS v1.3 since their following major and minor releases:

- [Java 11](https://docs.oracle.com/en/java/javase/11/security/java-secure-socket-extension-jsse-reference-guide.html)
- [Python 3.7](https://docs.python.org/3/library/ssl.html#tls-1-3)
- [Windows Server 2022](https://docs.microsoft.com/en-us/windows/win32/secauthn/protocols-in-tls-ssl--schannel-ssp-)
- [Apache 2.4.38 ](https://downloads.apache.org/httpd/CHANGES_2.4)
- etc

Would this be an alternative you would vote for @jmanico ?  If so then I'll modify this Pull Request.

jmanico:
I cannot endorse forcing TLS 1.3 at this juncture.

jmanico (commented):
I strongly oppose forcing the standard to TLS 1.2 - yet. Perhaps in another year.

cmlh:
@jmanico 

> While I get your point and think it’s fair - my vote on a TLS 1.3 only requirement today would be no. I’ll certainly vote yes to this in the future, but not now.

As @jmanico has confirmed this Pull Request is valid then it should be assigned to the future Milestone(s) following 4.1 rather than closed.

jmanico:
This is WAY out in the future, **this PR is staying closed**. But the issue itself can stay open. I do not want PR's lingering for years and this might no go live until 2024.

cmlh:
> This is WAY out in the future, **this PR is staying closed**. But the issue itself can stay open. I do not want PR's lingering for years and this might no go live until 2024.

I can't create another Pull Request from the same Git Commit after this has been closed by @jmanico

Issue #737, opened 2020-04-17
1.11.2 - [Clarification/For Discussion] for "unsynchronized state"


csfreak92:
ASVS 4.0 - 1.11.2 verification requirement is:

`Verify that all high-value business logic flows, including authentication, session management and access control, do not share unsynchronized state.`

I am confused about what the term "unsynchronized state" means here. There was no clarification or heading in ASVS documentation explaining this sub-section about this term. Can someone from the community or project leads please clarify what this means? And maybe we could add a sentence or two in this ASVS verification requirement as an edit to reflect it?

tghosth:
Good point, leaving this for when we prepare 4.1

csfreak92:
@tghosth, revisiting this issue since it still haunts our pentesters about what this means. Any idea about the "unsynchronized state" term here?

jmanico:
I suggest just delete this requirement, it makes no sense. Care to PR this @csfreak92 ? It needs to go!

elarlang:
I think we should be quite careful with deleting requirement in case we just don't clearly understand what it says - the risk is that requirement actually have good meaning but just bad wording.

In this case, linked CWE points to:
https://cwe.mitre.org/data/definitions/362.html

I can use this requirement for authentication when authentication state is not handled properly (it's not possible to verify is the same client reporting code for OAuth who actually started the authentication process) and therefore causing Race-Condition. Pure Race-Condition you can report to current 11.1.6, this one you can use for missing state. I feel I'm lacking English skills to explain that, so if it was not understandable then we can do another attempt.

jmanico:
The CWE helps me understand this more. Can you change "synchronized state" to "business logic workflow" so the issue is more clear? Otherwise, I'm +1 on a requirement for this topic. I get it now.

csfreak92:
Thanks for pointing that out @elarlang! That helped me understand this more. 

@jmanico, in that case I would reword this into: `Verify that all high-value business logic flows, including authentication, session management and access control, prevent race condition issues by not sharing business logic workflow` What do you think?

jmanico:
+1 I like this direction a lot. We’re getting close to a PR I’d enjoy merging to main.

One of the main attacks and I see in this area is step skipping attacks, like skipping the payment step in a e-commerce workflow. Can we someone emphasize non-security flows a bit better? Maybe?

--
Jim Manico
@Manicode

elarlang:
Just in case - for skipping steps there is requirement 11.1.1

[...]

PR #988, opened 2021-05-06
Update 0x10-V1-Architecture.md

csfreak92:
This is the proposal to improve the language for 1.11.2 and removing the term "unsynchronized state" to make it clearer. 

This Pull Request relates to issue #737


jmanico:
Can you drop the "it" for this one ?

_Verify that all high-value business logic flows, including authentication, session management and access control, maintain a consistent application state and check to prevent race conditions and business logic flaws._

And @elarlang do you support this PR?

elarlang:
and if we are here already... do we need "high-value"? what is high-value business logic flaw and what not that high value business logic flaw?

In general ok, we can clarify category and potential duplicate issues later.

jmanico:
Agreed Elar. So @csfreak92 would you care to change this to:

_Verify that business logic flows, including authentication, session management and access control, maintain a consistent application state and check to prevent race conditions and business logic flaws._

Issue #1002, opened 2021-05-19
Drop all requirements that discuss salting


jmanico:
ASVS should not be discussing salts at all.  All of the recommended password hashing algorithms salt for you.

https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

So all references salts should go away in 2.4.X and 2.4.2 should be removed

jmanico:
https://github.com/OWASP/ASVS/issues/1000

jmanico:
This has been merged

cmlh:
The merge should be performed by a different party to that of the requester.

PR #1016, opened 2021-06-03
Fixing (and reducing) references to cryptographic salts since approved password storage algorithms handle that for us automatically

jmanico:
Modification of several salt references


jmanico:
> Related to issue #1002.
> 
> > Password hashing functions take a password and one or more cost factors as inputs when generating a password hash.
> 
> This doesn't add much anymore. Should we remove it?

I removed this sentence from the PR. Looking ok to you?

Sjord:
Does 2.6.2 need a `[MODIFIED]` tag? Other than that, it looks ok.

jmanico:
I’ll fix that tonight!

cmlh:
I feel that this does not reflect the lessons learnt from trusting ESAPI (as one of many examples).

Unless the ASVS verification of the "Password Storage Algorithm" is published then this must be within the scope of ASVS and the ​"Password Storage Algorithm" should be verified against the published reference implementation [and associated supply chain].

Can I ask that this be reconsidered @jmanico please?

PR #1020, opened 2021-06-08
Remove 2.1.12

### LINKED ISSUE #570: V2.1.12 - Is this the duty of application developers or browsers?
> Verify that the user can choose to either temporarily view the masked entered password, or temporarily view the last typed character of the password.

I'm not sure that this is a fair requirement to put on application developers, if they use the password type for input fields containing passwords (which I think we want them to), then I think it's up to the browser to provide such controls unless I'm missing something. If not actionable by application developers, I would suggest to remove requirements they don't have influence over. If this advice is intended for non-browser-based applications, then I think we may need to clarify this so that web applications are not impacted.

Sjord:
Password input is the responsibility of the browser. Applications shouldn't
have to implement their own password entry control. NIST recommendations about
this are all optional.
The application shouldn't disable browser functionality that makes password
entry easy, but we already have 2.1.11 for that.

Closes #570.


elarlang:
Personally I don't mind to remove it, but there is some reason, why this requirement was built.

It is a bit larger question - how to be aligned with NIST. I recommend to carry out all argumentation in (reopened) issue and make PR after some "reaction time".

jmanico:
I agree this needs to go and will be accepting this PR. :)

Issue #1022, opened 2021-06-15
Use of non-✓ for level


Sjord:
The ASVS level columns generally contain a ✓. There are some exceptions to this, listed below. I propose to make these columns contain ✓ or nothing, and not free-form text. This makes it easier to retrieve a list of requirements for a certain level, and makes it clear which levels every requirement applies to.

| **2.8.7** | Verify that biometric authenticators are limited to use only as secondary factors in conjunction with either something you have and something you know. | | o | ✓ | 308 | 5.2.3 |
| **2.8.8** | [ADDED] Ensure that generation of the time-based multi-factor OTP token is based on the server's system time and not the client's machine. | | o | ✓ | 367 | 5.1.4.2 / 5.1.5.2 |
| **3.3.2** | If authenticators permit users to remain logged in, verify that re-authentication occurs periodically both when actively used or after an idle period. ([C6](https://owasp.org/www-project-proactive-controls/#div-numbering)) | 30 days | 12 hours or 30 minutes of inactivity, 2FA optional | 12 hours or 15 minutes of inactivity, with 2FA | 613 | 7.2 |
| **2.10.1** | Verify that intra-service secrets do not rely on unchanging credentials such as passwords, API keys or shared accounts with privileged access. | | OS assisted | HSM | 287 | 5.1.1.1 |
 None
| **2.10.2** | Verify that if passwords are required for service authentication, the service account used is not a default credential. (e.g. root/root or admin/admin are default in some services during installation). | | OS assisted | HSM | 255 | 5.1.1.1 |
| **2.10.3** | Verify that passwords are stored with sufficient protection to prevent offline recovery attacks, including local system access. | | OS assisted | HSM | 522 | 5.1.1.1 |
| **2.10.4** | Verify passwords, integrations with databases and third-party systems, seeds and internal secrets, and API keys are managed securely and not included in the source code or stored within source code repositories. Such storage SHOULD resist offline attacks. The use of a secure software key store (L1), hardware TPM, or an HSM (L3) is recommended for password storage. | | OS assisted | HSM | 798 | |

In 2.8.7 and 2.8.8, the level is "o". This means "Recommended, but not required", according to [this legend](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x11-V2-Authentication.md#legend). This is only used in two requirements, and the legend is specific to the authentication chapter. Furthermore, I think all requirements are recommended, so this doesn't add much information. I propose to remove the "o", and the legend chapter.

[...]

PR #1026, opened 2021-06-17
Remove specifics from level columns in V3.3 Session Termination

Sjord:
Related to #1022.

### DIFF
--- a/4.0/en/0x12-V3-Session-management.md
+++ b/4.0/en/0x12-V3-Session-management.md
@@ -30,14 +30,12 @@ TLS or another secure transport channel is mandatory for session management. Thi
 
 ## V3.3 Session Termination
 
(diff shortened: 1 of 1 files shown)


elarlang:
With this one I think ASVS should be independent document - if something is taken from NIST and it's part of requirement, then document should duplicate it.

Requirements should be understandable without opening any external document (like NIST in this example).

Removed "30 days | 12 hours or 30 minutes of inactivity, 2FA optional | 12 hours or 15 minutes of inactivity, with 2FA" from requirement must be still covered in ASVS.

jmanico:
I hate to see this info go away. Can we move it somewhere else in the doc?

Sjord:
I added a [table with specifics](https://github.com/OWASP/ASVS/blob/2935cf7d898c06297d3c459e8d86ca494cf27a3b/4.0/en/0x12-V3-Session-management.md#v33-session-termination). What do you think?

elarlang:
I think it makes sense to open separate issue for the discussions on that topic. Discussion over pull-requests is a bit messy.

Provided table gives nice overview, but it makes requirement text itself dependent on document. Maybe there are some other proposals or ways.

Sjord:
Please continue discussion in #1022.

Issue #1066, opened 2021-10-12
14.2.5 Verify that an inventory catalog is maintained of all third party libraries in use.


Sjord:
[ASVS/0x22-V14-Config.md at master · OWASP/ASVS](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x22-V14-Config.md#v142-dependency)

> 14.2.5 Verify that an inventory catalog is maintained of all third party libraries in use.

What is the goal of this? Isn't this just the package-lock.json in the project? What is this inventory catalog meant for? If the inventory catalog is maintained, but never used for anything, it doesn't really help.

coderpatros:
> Isn't this just the package-lock.json in the project?

For a lot of ecosystems the package manifest is not necessarily what will be resolved and included in the build. Especially if the version in the manifest specifies a version constraint as opposed to a specific version.

There's also an ecosystem where the package lock file information is only enforced when restoring packages, not when resolving packages during the build.

jmanico:
I really do not want to keep long term issues dangling. Can we please leave this closed for now since the original issue is resolved?

At the very least, if you want something new added let’s open a new issue and track these things discretely.

Is this acceptable?

coderpatros:
> We'll also need to settle the following proposals too?

I probably wouldn't focus on specific formats (as much as I'd love a reference to CycloneDX). But rather what information is being included in the SBOM. It might make sense to just reference the V2 L1-3 requirements in SCVS https://owasp-scvs.gitbook.io/scvs/v2-software-bill-of-materials

jmanico:
Patrick,

Would you care to give a PR a go for this? Or give us a suggestion as to 
the enhanced text for this requirement?

PR #1069, opened 2021-10-13
Replace "inventory catalog" with Software Bill of Materials (SBOM)

cmlh:
Fix #1066 (credit to @Sjord)

@jmanico approval at https://github.com/OWASP/ASVS/issues/1066#issuecomment-941636201

PR #1074, opened 2021-10-15
Insert [MODIFIED] for @elarlang

cmlh:
Fix https://github.com/OWASP/ASVS/issues/1066, specifically https://github.com/OWASP/ASVS/pull/1069#issuecomment-944531094


jmanico:
Already got this but thank you Christian!

PR #1080, opened 2021-10-16
DRAFT: Create 5.0 directory

### LINKED ISSUE #1079: Create 5.0 directory
A 5.0 directory should be created to hold new work, and 4.0 should reflect the state of the last release.

ike:
Fixes #1079.

This is a draft. The 4.0 directory MUST be validated as representing the 4.0.2 release.


elarlang:
I don't the reason why we should have anything else than "en/" folder for content for 5.0.

Eg, in my opinion we don't need any "docs_*" folder and "es/", "fr/" folders.

The same way, no need for old documents. Those should belong to 4.0/ folder, or just folder above but not in 5.0 folder.

One more idea - as folder 5.0 seems really "official", should we have folder named "5.0-draft/" instead of 5.0 (or it's enough to use some big-red-text "hey you, it's bleeding edge!"?

For confirming - there is no need for extra eyes, there just need to write some "diff" from different checkout :)

ike:
Great suggestion -- I will remove everything from 5.0 except for the automation tooling and the `en` directory.

I double-checked my own diffing and found some cruft -- just force-pushed an updated commit where the only change in the diff is the updated `.gitignore`

ike:
asvs-diff-10-16-21

jmanico:
I think all of Elar’s suggestions are great!

jmanico:
Also, we are about to merge in a Spanish translation that we want to be part of the 4.02 release. So when it's time to roll back the 4.0 folder to the 4.02 release we need to account for that in some way. We can also just roll back to the real 4.02 cutout point and then add in the new translation if that is easier.

ike:
If possible, we should wait and merge this PR first. The Spanish translation PR can have changes for both the `4.0` and `5.0` directories.

jmanico:
Which one is this?

ike:
Edited for clarity -- "this" is the `Create 5.0 Directory` PR.

PR #1082, opened 2021-10-19
label and md syntax corrections

elarlang:
label and md syntax corrections


jmanico:
I just reviewed all of these changes. Thanks for cleaning these mistakes up!

PR #1120, opened 2021-10-29
Meta-makefile for compilation

### LINKED ISSUE #1078: Dockerfile entrypoint should be script
The Dockerfile entrypoint should be a script that executes `make` and then, if successful, executes `make clean`. This will allow us to always clean up after a successful compilation, without breaking makefile conventions.

ike:
Fixes #1078.

This change adds a new Makefile to the root directory, which contains `latest`, `all`, `4.0`, and `5.0` targets. It also detects whether you have the docker image built, and if you don't it builds it for you.

The entrypoint for the docker image is now `docker/run.sh`, which allows us to configure parameters per version, if necessary, and also runs `make clean` after compilation.

All you have to do to build bleeding edge now is run `make` in the root directory. Further instructions are outlined in `COMPILING.md`.


ike:
Would like @lfservin to take a look at this too.

jmanico:
I'm tempted to just merge, but I'll wait until others can review.

cc @tghosth

ike:
I am planning on making another PR that will address docker image size.

lfservin (commented):
We probably need both  a BUILDDIR and a DISTDIR. First we parse the vanilla markdown files from the repository and enhance them with hyperlinks (lines 40,41) then we use these md files to generate the pdf (where we want the hyperlinks to exist) but not for generating json or the other formats. 
By putting everything together, the dist directory would include the modified markdown files.

lfservin:
> 
> 
> I am planning on making another PR that will address docker image size.

The Docker file image is as small as I could get it. Unfortunately latex is huge. If you look into the Dockerfile, only the most essential base latex is installed. Then only the packages that are absolutely needed for this document were added one by one.

Now the base image is a whole ubuntu. we could switch to alpine and reduce the size. I'll give that a try and make a PR

ike:
I switched out for alpine locally, and it saved about 100mb, which was less than I expected. I can push that to another branch if you want to experiment with it?

lfservin:
> 
> 
> I switched out for alpine locally, and it saved about 100mb, which was less than I expected. I can push that to another branch if you want to experiment with it?

May I refer you to this stack exchange answer: https://tex.stackexchange.com/a/493882 the difference in the base images alpine vs ubuntu turns out not to be too big, compared to all the weight that latex brings with it.

Focal brings 27MB as a base image. Alpine brings almost 3 MB as a base image. The 20-something MB difference is negligible when looking at the final image size (~1GB)

For latex, pandoc (and thus python) to work we need: build tools, fonts, perl, python, lua. Then we install tinytex and only those latex packages needed (using tlmgr) to build. A couple of fonts are added, but they don't bring too much weight with them. 

I haven't found a lighter-weight latex installation, unfortunately. I spent some time researching alternatives and came thus to the current state.

ike:
From my analysis, about 400MB appears to be font files. Could we limit the fonts that are installed?

jmanico:
That’s a lot of fonts! I am guessing we do not use them all.

lfservin (commented):
I'm not sure whether this is the right way to generate the meta-makefile. In doing so, we deny the possibility of passing targets to build as parameters to the docker container. E.g. 
``` 
docker run -v path/:data --rm asvs/documentBuilder json
``` 
This would only build the target "json"

The makefile works well in both versions 4 and 5. The source directory (SOURCEDIR) could be made to depend e.g. upon an environment variable. Then the Path could be adjusted accordingly. E.g.
``` 
docker run -v path/:data --rm  -e VERSION=5.0 asvs/documentbuilder
``` 
This would create an environment variable "VERSION" with a value "5.0". Now the directories are conveniently named 4.0 and 5.0 and the structure is pretty much the same. 

## Proposal
Let the makefile build the path to the target. Default would be 5.0 but can be changed to 4.0. A variable named "BASEDIR" could be made to depend on the environment variable "VERSION". Then the path to the corresponding sources and templates is built dynamically:
```
BASEDIR=$pwd/$(VERSION)
``` 
for this to work the `Makefile`  would need to be a level down in the directory structure, i.e. outside of the version. Likewise the directories `tools` and `docker`, since AFAIK there are no breaking changes between the MD files in v4 and v5.

This would make docker/run.sh unnecessary. It would all be handled by the one and only Makefile.

lfservin:
> 
> That’s a lot of fonts! I am guessing we do not use them all.

I just built the container and I'm trying to figure out how much space is taken by fonts. This is the result I got:
image

The fonts folder shows 113 MB. We need noto for the japanese "A" character. Adobe fonts for the types in use in the document. Glad to use others.

lfservin:
> 
> 
> > That’s a lot of fonts! I am guessing we do not use them all.
> 
> I just built the container and I'm trying to figure out how much space is taken by fonts. This is the result I got: image
> 
> The fonts folder shows 113 MB. We need noto for the japanese "A" character. Adobe fonts for the types in use in the document. Glad to use others.

TinyTex is 475 MB
Pandoc brings about 100 MB

lfservin:
> 
> 
> From my analysis, about 400MB appears to be font files. Could we limit the fonts that are installed?
> 
> I haven't figured out what is installing the fonts, but there's a lot of them.

Adobe fonts are downloaded in the stage called "font_base". There's a CURL to get them from github. We're downloading these fonts: source-sans-pro, source-serif-pro, source-code-pro
other fonts packages are : fonts-noto-cjk and libfreetype6. These are installed in the base stage.

ike:
> This would make docker/run.sh unnecessary. It would all be handled by the one and only Makefile.

What about passing formats in to the docker container? I put an example here: https://github.com/OWASP/ASVS/pull/1120/commits/6b2c61f06d15d61c626376438bcb6e68696c6915

With this change, you'd target a particular format by calling:

```
make 5.0 FORMATS="csv json"
```

ike:
> TinyTex is 475 MB
> Pandoc brings about 100 MB

Ahhhh yes, I found a font folder in `.TinyTex/texmf-dist` which is what brought the font total to ~350-400mb.

Screenshot_2021-11-04_21-58-54

ike:
> We probably need both a BUILDDIR and a DISTDIR. First we parse the vanilla markdown files from the repository and enhance them with hyperlinks (lines 40,41) then we use these md files to generate the pdf (where we want the hyperlinks to exist) but not for generating json or the other formats.

Yes, this is good -- I also fixed the `make clean` so the copied files are deleted now, so that's another way we could solve it. I am fine having a temporary "build" dir and a final compiled dir too.

lfservin (commented):
No need for the second environment variable. FORMATS is not needed. That can be solved natively by Docker.  The ENTRYPOINT is "make" which means that any commands "CMD" you want to pass are just given after the container name. Thus it would look like this
```
docker run --rm -v "`pwd`/5.0:/data" -v "`pwd`/docker:/scripts" -e "ASVS_VERSION=5.0" asvs/documentbuilder target1 target2 target3
```
Where targetN is any of the targets in the Makefile, i.e. pdf, json, csv, xml, json-flat

We don't need docker/run.sh. The makefile is making both versions 4 ad 5.

ike:
I am trying to avoid making people remember how to correctly run the docker command -- using the nested makefiles allows us to do that, but we could also use a shell script or similar.

Do you have preferences on how to keep the command for end users simple?

ike:
> ```docker run --rm -v "`pwd`/5.0:/data" -v "`pwd`/docker:/scripts" -e "ASVS_VERSION=5.0" asvs/documentbuilder target1 target2 target3```

Another way to put it -- yes, this command is totally right, but I want to expose a simpler interface to the user, i.e. `build-command-thingy 5.0 target1 target2 target3`. I don't really care how `build-command-thingy` is constructed, but it should be extensible and flexible for future changes.

I think the `make 5.0 FORMATS="target1 target2 target3"` fits this bill, but I am curious if you have other suggestions of how to simplify this interface.

lfservin:
Could we synch over slack, Twitter,  or may i propose a short call? In both
Twitter and the OWASP slack you can find me as @lfservin we can exchange pm
to exchange emails and setup the call.

ike:
That sounds great! I'll hit you up on Monday.

jmanico:
Thank you both very much for working on our automation. Ike, if you are ever waiting for me to merge let me know. I'm hesitant to merge automation PR's until you tell me to. :)

ike:
Alright, I attempted to remove all the extra fonts from the TinyTex installation in the dockerfile, but now PDFs won't generate. I will have to be more surgical I think -- I'll keep y'all apprised.

jmanico:
Is this ready for merging? Hey @ike :)

ike:
👍 There's more to come, but this does accomplish a lot of our goals. I am still working on an HTML templating prototype, but it'll use this same architecture.

lfservin:
Requirement [14.6.3](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x22-V14-Config.md#v146-http2) adds these characters (\r\n), which make need to be escaped in latex. 
I just made a pull request #1294 which fixes this and improves on the Makefile in the version 5.0

PR #1126, opened 2021-11-07
8.2.2 improvement, closes #1091

### LINKED ISSUE #1091: 8.2.2 - "with the exception of session tokens" is a bit too wide
Link to bleeding edge: [8.2.2](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x16-V8-Data-Protection.md#v82-client-side-data-protection)

"spin-off" requirement from The Monster Issue https://github.com/OWASP/ASVS/issues/843#issuecomment-883613794

Current (_bleeding edge_) solution
> 8.2.2 Verify that data stored in browser storage (such as localStorage, sessionStorage, IndexedDB, or cookies) does not contain sensitive data, with the exception of session tokens.

v4.0.2-8.2.2
> Verify that data stored in browser storage (such as HTML5 local storage, session storage, IndexedDB, or cookies) does not contain sensitive data or PII.

v4.0.2-3.2.3 (deleted from _bleeding edge_):
> Verify the application only stores session tokens in the browser using secure methods such as appropriately secured cookies (see section 3.4) or HTML 5 session storage.

For cookie-based sessions there is now separate requirement 3.4.6 in _bleeding edge_:
> 3.4.6 Verify that cookie-based session tokens are only transferred in Set-Cookie and Cookie headers.

Change made for for v4.0.3 (just drop the exception, because for v4.0.3 3.2.3 stays and there is no 3.4.6):
> 8.2.2 Verify that data stored in browser storage (such as localStorage, sessionStorage, IndexedDB, or cookies) does not contain sensitive data.

**Problem - a bit too wide open exception**

Session token IS sensitive piece of information AND it is stored in some browser storage and therefore anyway in conflict with 8.2.2 (without added exception):

  * cookies for cookie-based sessions
  * sessionStorage for token-based sessions

With wide-open exception _with the exception of session tokens_ it is allowed to store any type of token in any browser storage.

Solution 1) Should we be more precise with this exception? like (wording needs improvement):

> , with the exception of cookie-based session tokens in cookies and token-based session tokens in sessionStorage.

Solution 2) Should we add extra requirement to V3.5.* which says something like:
> Verify the application only stores token-based session tokens in the browser sessionStorage

Then previous 3.2.3 is covered with new 3.4.6 and new 3.5.*

ping @jsulinski

elarlang:
This Pull Request relates to issue #1091

### DIFF
--- a/5.0/en/0x16-V8-Data-Protection.md
+++ b/5.0/en/0x16-V8-Data-Protection.md
@@ -28,7 +28,7 @@ Ensure that a verified application satisfies the following high level data prote
 | # | Description | L1 | L2 | L3 | CWE |
 | :---: | :--- | :---: | :---:| :---: | :---: |
 | **8.2.1** | Verify the application sets sufficient anti-caching headers so that sensitive data is not cached in modern browsers. | ✓ | ✓ | ✓ | 525 |
-| **8.2.2** | [MODIFIED] Verify that data stored in browser storage (such as localStorage, sessionStorage, IndexedDB, or cookies) does not contain sensitive data, with the exception of session tokens. | ✓ | ✓ | ✓ | 922 |
+| **8.2.2** | [MODIFIED] Verify that data stored in browser storage (such as localStorage, sessionStorage, IndexedDB, or cookies) does not contain sensitive data, with the exception of cookie-based session tokens in cookies and token-based session tokens in sessionStorage. | ✓ | ✓ | ✓ | 922 |
 | **8.2.3** | Verify that authenticated data is cleared from client storage, such as the browser DOM, after the client or session is terminated. | ✓ | ✓ | ✓ | 922 |
 
 ## V8.3 Sensitive Private Data


jmanico:
Can we clean this up just a little before merging?

From:

Verify that data stored in browser storage (such as localStorage, sessionStorage, IndexedDB, or cookies) does not contain sensitive data, with the exception of cookie-based session tokens in cookies and token-based session tokens in sessionStorage.

to

Verify that data stored in browser storage (such as localStorage, sessionStorage, IndexedDB, or cookies) does not contain sensitive data, with the exception of cookie-based session tokens and token-based session tokens in sessionStorage.

elarlang:
Do you wnat to change the meaning to "cookie-based session tokens are allowed to be stored in sessionStorage"? 

This is not true and this is not the goal for/from me.

jmanico:
All I am saying is that the phrase“ cookie-based session tokens in cookies” is redundant.

elarlang:
I don't agree, without this phrase the meaning is different. If we can not agree on this or it's not clear like it is, we just need to drop this PR and come with new solution.

jmanico:
Where else would a cookie-based session be stored in a browser? I am confused...

It is **_implicit_** that a "cookie-based session" is stored in a cookie.

jmanico:
Hey @tghosth and/or @danielcuthbert can you please take a look at this?

elarlang:
> Where else would a cookie-based session be stored in a browser? I am confused...
> 
> It is **_implicit_** that a "cookie-based session" is stored in a cookie.

And this is precisely the reason why we need to make it clear - that for cookie-based session token it is allowed to store the token ONLY in cookies, and do not send it to DOM, store in sessionStorage etc.

I added requirement V3.4.6 and without saying it/duplicate again in 8.2.2, it is in conflict with 3.4.6. 

If needed, please go through all the discussion in monster issue which lead me to make this proposal:
https://github.com/OWASP/ASVS/issues/843#issuecomment-798791597

jmanico:
I went through all of the threads. I agree with with thew spirit of this 
requirement. My concern is that the use the term /cookie/ twice in a 
sentence like this is just bad English and it reads poorly. I'd like 
other leads to chime in before we merge this.

Issue #1137, opened 2021-11-19
Chapter name parsing is broken


Gby56:
Hi,

This is just a simple issue I found, the csv and other outputs don't have the Chapter Name properly filled out, the regex was broken apparently, because the markdown input (in 4.0.x) isn't using `:` character
I have a PR ready for this already :)

PR #1138, opened 2021-11-19
Fix Chapter name parsing with regex

Gby56:
This Pull Request relates to issue #1137
The regex was broken, fixed it !
Glad I could help ! I tested it with https://regex101.com/ and on my personal machine it ran perfectly

### DIFF
--- a/4.0/asvs.py
+++ b/4.0/asvs.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-
+# -*- coding: utf-8 -*-
 ''' ASVS document parser and converter class.
 
     Based upon code written for MASVS By Bernhard Mueller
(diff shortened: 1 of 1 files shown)


jmanico:
Hey @ike I merged this but can you make sure I did not do so in haste?

ike:
I haven't tested it locally but visual review looks good! If I notice a problem I will report back.

Gby56:
I was surprised there was no Github action triggering an automatic export for testing, just to see if it works properly (I dev on macOS, you never know...)
Could that be interesting for you ? I can take a look

Gby56:
Also I was wondering, will the current 4.0.3 release be updated ? Since the CSV is missing the data.

ike:
> I was surprised there was no Github action triggering an automatic export

We're working on the tooling to support this, right now we don't have automated builds. Feel free to check out the Markdown label for where we're currently at.

Issue #1153, opened 2021-12-16
V14.5.6 Refer to RFC


svdb-ncc:
[14.5.6](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x22-V14-Config.md#14.5) mentions the conflicting Transfer-encoding and Content-Length headers. The wording is still open to interpretation while if we would say that RFC 2616 (in which is says: "If a message is received with both a Transfer-Encoding header field and a Content-Length header field, the latter MUST be ignored.") should be followed correctly, we more specific.

PR #1155, opened 2021-12-16
Refer to RFC 2616

svdb-ncc:
See #1153


elarlang:
We don't use markdown syntax in requirement text

jmanico (commented):
Can you kindly go with this instead? Just a little wordsmithing.

| **14.5.6** | [ADDED] Verify that the infrastructure follows RFC 2616 and ignores the Content-Length header field if a Transfer-Encoding header field is also present. | | ✓ | ✓ | 44

svdb-ncc:
Made the changes as suggested

Issue #1154, opened 2021-12-16
HTTP/2 is not addressed


svdb-ncc:
HTTP/2 is a relatively new binary protocol rather than text-based.
There are a number of vectors that are possible, especially when downgrading is performed. These should be included here in the ASVS.

Some training material regarding this from PortSwigger Web Security Academy:
* [HTTP/2-exclusive vectors](https://portswigger.net/web-security/request-smuggling/advanced/http2-exclusive-vectors)
* [HTTP/2 downgrading](https://portswigger.net/web-security/request-smuggling/advanced/http2-downgrading)

elarlang:
| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **14.6.1** | [ADDED] Verify that the value in the Content-Length matches the calculated length using the built-in mechanism. | ✓ | ✓ | ✓ | 400 |

For me it seems like some missing word or shortening here: "Verify that the value in the Content-Length matches..."

Proposals:
* "Verify that the value in the Content-Length **header** matches..."
* OR "Verify that the value in the Content-Length **response header** matches..."

jmanico:
I like this suggestion. Perhaps "Content-Length response header matches..."

svdb-ncc:
I will change it to request header

elarlang:
Requirements got initially in via #1154 / #1156 https://github.com/OWASP/ASVS/commit/367d6feb812d072c2837293066230c8b8f894b4b

I reopen, as I think those requirements needs some explanation into those that is clearly understandable, which vector those mitigate or why they exists.

Currently those are located in "V13.7 HTTP/2"

| # | Description | Level | CWE |
| :---: | :--- | :---: | :---: |
| **13.7.1** | [ADDED] Verify that the value in the Content-Length request header field matches the calculated length using the built-in mechanism. | 1 | 400 |
| **13.7.2** | [ADDED] Verify that all Transfer-Encoding header fields are stripped from the message or that the request is blocked entirely. | 1 | |
| **13.7.3** | [ADDED] Verify that a full CRLF (\r\n) sequence is neutralized inside a HTTP/2 header. | 1 | 113 |

jmanico:
How about:

| # | Description | Level | CWE |
| :---: | :--- | :---: | :---: |
[...]

PR #1156, opened 2021-12-16
HTTP2

svdb-ncc:
This Pull Request relates to issue #1154


jmanico:
Very astute RP, thank you, Steven!

elarlang:
We don't use markdown syntax in requirement text.

PR #1173, opened 2022-01-27
Update 0x01-Frontispiece.md

aref2008:
Add Arabic translation team to contributors list.


elarlang:
Hi Aref (ping @aref2008 )!

Thank you for your effort! What we would like to see, is translation in markdown format and then we can generate pdf and other formats out of that.

To follow the progress, I opened separate issue for that: https://github.com/OWASP/ASVS/issues/1174

elarlang:
About this PR - translator get mentioned in [repo README](https://github.com/OWASP/ASVS#translations) and I try to find solution, that translators are covered in translated document as well.

The reason here is simple - we can not make change to already released v4.0.3 data, because it's released and then we should create new release.

Issue #868, opened 2020-10-26
Better clarify the difference between ASVS and SAMM


omerlh:
Context: [Twitter](https://twitter.com/omerlh/status/1320425822310772736?s=20)

It seems there is some overlapping between SAMM and ASVS (V1.1, V14, prefix). I think it could help further readers to add something around the relationship between SAMM and ASVS and how it is best to use them both.

tghosth:
I think this is closed by #1213 and #1210, @omerlh please let me know if you think this is still lacking.

PR #1210, opened 2022-02-02
Remove diagram that I do not understand

tghosth:
I thought about this based on issue #868. I don't think this PR resolves that issue but it is kinda of a start.

I don't understand this diagram because it seems to confuse ASVS and secure development lifecycle (which would be more suited to SAMM). I think that the point is that the ASVS can be used at lots of stages but I am not sure that is how it reads at the moment. I think we should remove the diagram.

I would also like to make the difference clearer in the text but that will come later.


elarlang:
If we talk about 5.0 perspective - from some moment I prefer to use "white paper" and then put back only those things which are valid.

Before we need to have defined the scope, levels, concept etc. If those are done, I would say all the text part, including diagrams, are not valid anymore.

tghosth:
As a principle, I don't think we can edit a "blank page". In general, I think we need to start from our existing basis. 

However, I don't think we will get to a stage where this diagram can be edited to make more sense, nor do we have the source which is why I think. I have a seen at least one case where this has caused confusion and in parallel I am working to improve the overall description of the ASVS.

As such, as part of this, I think we should remove the diagram. If you agree, I need you to approve :)

elarlang (approved):

Issue #1038, opened 2021-07-24
JWT Secure best practices


csfreak92:
Do we have a requirement for JWT secure best practices such as checking of algorithm (definitely should not be set to none), ensuring expiration of tokens, checking of issuer, etc? I keep looking for them but it seems the closest one we have was mentioned in #839 and it was pertaining to requirement 3.5.3 - `Verify that stateless session tokens make use of digital signatures to protect against tampering.` 

There is nothing specific for JWTs, so maybe this is a good idea for developers to have in mind as part of their secure coding checklist. I've seen applications using JWTs but tokens are not being used properly as with expiration, decoded claims contain information that can be modified and result to gaining admin access or cross-user access (IDORs), bypass authentication mechanisms, etc. If we have anything else, then maybe I was just mistaken in reading most of the ASVS requirements.

csfreak92:
Placing this RFC here for reference: https://datatracker.ietf.org/doc/html/rfc8725

jmanico:
839 is all we have and I would love to see more. Bring it on!

jmanico:
Would you care to help us convert this to requirements?

csfreak92:
Definitely! I will place it here as comments first before submitting a PR.

cmlh:
Can [RFC8725](https://datatracker.ietf.org/doc/html/rfc8725) be forked from ASVS as JWT is specific to JSON (unless [RFC8725](https://datatracker.ietf.org/doc/html/rfc8725) can be made generic and therefore applicable to all technologies)?

csfreak92:
Sounds like a great idea, let me draft it and place it here, thanks for pointing that out @cmlh!

csfreak92:
Draft/Proposal for JWT secure best practices:  

`1) Verify expiration of JWTs are checked in the backend service (e.g. exp - expiration time and iat - issued before current time). `
`2) Verify that algorithm validation is being done by the backend service for the JWTs used and as such it should not accept "None" value in the algorithm section. `
`3) Verify proper validation of the JWT payload claims are done by the backend service (e.g. iss - valid issuer, sub - subject, aud - audience, etc). `
`4) Verify that signatures of JWTs are present in the claims to add more protection from tampering. (JWS-JSON Web Signature) `

[...]

PR #1217, opened 2022-02-08
Enhance stateless session token wording

tghosth:
This Pull Request relates to issue #1038. 

@elarlang do you have further questions?
@csfreak92 and @jmanico do you have comments?

### DIFF
--- a/5.0/en/0x12-V3-Session-management.md
+++ b/5.0/en/0x12-V3-Session-management.md
@@ -15,7 +15,8 @@ As previously noted, these requirements have been adapted to be a compliant subs
 
 | # | Description | L1 | L2 | L3 | CWE | [NIST &sect;](https://pages.nist.gov/800-63-3/sp800-63b.html) |
 | :---: | :--- | :---: | :---: | :---: | :---: | :---: |
-| **3.1.1** | Verify the application never reveals session tokens in URL parameters. | ✓ | ✓ | ✓ | 598 | |
+| **3.1.1** | Verify that the application never reveals session tokens in URL parameters. | ✓ | ✓ | ✓ | 598 | |
+| **3.1.1** | Verify that the application performs all session tokens verification using a trusted, back-end service. | ✓ | ✓ | ✓ | 603 | |
 
 ## V3.2 Session Binding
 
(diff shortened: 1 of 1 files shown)


csfreak92:
@tghosth, the wording reads better now. Although question on `"| **3.5.5** | [ADDED] Verify that the signature of a stateless token is being checked before processing it further and the algorithms is checked against an allow-list. | ✓ | ✓ | ✓ | 347 | |"`, for better implementations of a stateless token signature validation, would that be considered an allow-list (whitelist) or could there be another case? I guess an allow-list is more flexible than a config that checks for this?

tghosth:
I think we would only ever want an allow-list of acceptable algorithms, I cannot think of another suitable option. Note that I also made some changes to the wording to make things clearer :)

5.0/en/0x12-V3-Session-management.md
```diff
@@ -16,7 +16,7 @@ As previously noted, these requirements have been adapted to be a compliant subs
 | # | Description | L1 | L2 | L3 | CWE | [NIST &sect;](https://pages.nist.gov/800-63-3/sp800-63b.html) |
 | :---: | :--- | :---: | :---: | :---: | :---: | :---: |
 | **3.1.1** | Verify that the application never reveals session tokens in URL parameters. | ✓ | ✓ | ✓ | 598 | |
-| **3.1.1** | Verify that the application performs all session tokens verification using a trusted, back-end service. | ✓ | ✓ | ✓ | 603 | |
+| **3.1.1** | [ADDED] Verify that the application performs all session tokens verification using a trusted, back-end service. | ✓ | ✓ | ✓ | 603 | |
```
elarlang, line 19: there seems to be some mistake.

If we "replace" requirement, it makes sense to mark one requirement [DELETED, + reason] and new requirement must be with new unique identifier with [ADDED] label

tghosth: OK, does this look better now @elarlang:
https://github.com/OWASP/ASVS/pull/1217/commits/d4f097d47fb1c7fce7bf5df3e9196d30527a94f5
https://github.com/OWASP/ASVS/pull/1217/files

elarlang (approved):

Issue #999, opened 2021-05-19
2.2.3


jmanico:
2.2.3 and 2.2.7 overlap some.

I suggest we drop the second sentence from 2.2.3. The second sentence is messy and adds confusion here and I think it should go.

Verify that secure notifications are sent to users after updates to authentication details, such as credential resets, email or address changes, logging in from unknown or risky locations. 

**The use of push notifications - rather than SMS or email - is preferred, but in the absence of push notifications, SMS or email is acceptable as long as no sensitive information is disclosed in the notification.**

Sjord:
> [2.2.3](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x11-V2-Authentication.md#:~:text=verify%20that%20secure%20notifications%20are%20sent%20to%20users%20after%20updates%20to%20authentication%20details%2C%20such%20as%20credential%20resets%2C%20email%20or%20address%20changes%2C%20logging%20in%20from%20unknown%20or%20risky%20locations.%20) Verify that secure notifications are sent to users after updates to authentication details, such as credential resets, email or address changes, logging in from unknown or risky locations. 

I would propose to split this in two:
1. notifications are sent to users after updates to authentication details, such as credential resets, email or address changes
2. logging in from unknown or risky locations, and/or [verify intent to authenticate by requiring](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x11-V2-Authentication.md#:~:text=verify%20intent%20to%20authenticate%20by%20requiring).

I think 1 is quite easy to implement and could be L1, whereas 2 is a bit harder and could be L2 or L3.

jmanico:
I support splitting these into two requirements as suggested above.

Sjord:
I propose:

> 2.2.3: Verify that users are notified after updates to authentication details, such as credential resets, or modification of the username or email address.

I removed the requirement for "secure notifications", because SMS and email are also acceptable. There is no confidentiality requirement for a message that your password changed.

For notification on authentication attempts, I would propose something like this:

> 2.2.8: Verify that users are notified of suspicious authentication attempts, even when (partially) successful. Suspicious authentication attempts include authentication from an unusual location or client, correct authentication with only one of multiple factors, and correct authentication after a long period of inactivity.

Initially I planned to merge that with 2.2.7:

> 2.2.7: Verify intent to authenticate by requiring the entry of an OTP token or user-initiated action such as a button press on a FIDO hardware key.

[...]

PR #1240, opened 2022-03-01
Split 2.2.3

Sjord:
Notify on credential changes, and notify on "suspicious" authentication attempts.

Remove CWE 620, as that is about requiring the old password on password change, not about notification.

Couple to NIST 6.2.1, as that has "CSP SHOULD send a notification" a couple of times.

Closes issue #999.


5.0/en/0x11-V2-Authentication.md
```diff
@@ -59,13 +59,14 @@ NIST considers SMS as ["restricted" authenticator types](https://pages.nist.gov/
...
 | **2.2.6** | Verify replay resistance through the mandated use of One-time Passwords (OTP) devices, cryptographic authenticators, or lookup codes. | | | ✓ | 308 | 5.2.8 |
 | **2.2.7** | [DELETED, DUPLICATE OF 2.3.2] | | | | | |
 | **2.2.8** | [ADDED] Verify that all failed authentication challenges respond in the same average response time. | | ✓ | ✓ | | |
 | **2.2.9** | [ADDED, SPLIT FROM 2.2.4] Verify that multi-factor authentication is required, that is, the application uses either a multi-factor authenticator or a combination of single-factor authenticators. | | ✓ | ✓ | 308 | 4.2.1 |
+| **2.2.10** | [ADDED, SPLIT FROM 2.2.3] Verify that users are notified of suspicious authentication attempts. Suspicious authentication attempts may include successful or unsuccessful authentication from an unusual location or client, partially successful authentication with only one of multiple factors, successful or unsuccessful authentication after a long period of inactivity or successful authentication after several unsuccessful attempts. | | ✓ | ✓ | 778  | |
```
tghosth: Please can you confirm that there is no similar requirement to this notification requirement anywhere else in the standard, for example in the logging section.

5.0/en/0x11-V2-Authentication.md
```diff
@@ -59,13 +59,14 @@ NIST considers SMS as ["restricted" authenticator types](https://pages.nist.gov/
...
 | **2.2.2** | [MODIFIED] Verify that weak authenticators (such as out-of-band authentication using VOIP or email) are not used as authentication methods. Verify that restricted authenticators (those using PSTN to deliver OTPs via phone or SMS) are offered only when alternate stronger methods are also offered and when the service provides information on their security risks to users. | ✓ | ✓ | ✓ | 304 | 5.2.10 |
-| **2.2.3** | Verify that secure notifications are sent to users after updates to authentication details, such as credential resets, email or address changes, logging in from unknown or risky locations. The use of push notifications - rather than SMS or email - is preferred, but in the absence of push notifications, SMS or email is acceptable as long as no sensitive information is disclosed in the notification. | ✓ | ✓ | ✓ | 620 | |
+| **2.2.3** | [MODIFIED] Verify that users are notified after updates to authentication details, such as credential resets or modification of the username or email address. | ✓ | ✓ | ✓ | 778 | 6.1.2 |
```
tghosth, line 62: I made some changes to both requirements including adding [CWE 778](https://cwe.mitre.org/data/definitions/778.html) and rewording 2.2.10. Are you comfortable with these changes?

Sjord: CWE 778 is really a stretch, since that is about logging and not user notifications.

5.0/en/0x11-V2-Authentication.md
Sjord: Related requirements are:

* 2.5.5 Verify that if an authentication factor is changed or replaced, that the user is notified of this event.
* 2.8.5 Verify that if a time-based multi-factor OTP token is re-used during the validity period, it is logged and rejected with secure notifications being sent to the holder of the device.

5.0/en/0x11-V2-Authentication.md
jmanico: @Sjord can you give me an edit that illustrates what you want to see here? I think we are close to merging this. I agree making the requirement less wordy would be prudent.

Sjord: I propose:

> Verify that users are notified of suspicious authentication attempts. Suspicious authentication attempts include attempts from an unusual location or client, after a long period of inactivity, and partially successful authentication with only one of multiple factors.

I left out "successful authentication after several unsuccessful attempts". I think users should be notified when their account is blocked because of too many login attempts, but perhaps that should not be in this requirement. I don't think this notification should only be done after successful authentication.

tghosth: @jmanico any feedback on @Sjord's response?

jmanico (approved):

Issue #1256, opened 2022-04-06
Several Minor Typographical Errors in 4.0.3 and 5


ussjoin:
Hello! While going through the ASVS control objectives, I came across several extremely minor typographical and/or grammatical errors in some sections of text. To assist the project, I am creating a pull request (linked to this issue) with changes in both the version 4 head and version 5 head Markdown files, as the text in each case was identical in 4 and 5.

In one case, I made a parallel change in the Arabic translation, as that translation includes some English words which included the typo.

In no cases are there any substantive changes to the intention of the text; these are just to help clean up the text.

The sections in the associated PR are:
V1-Architecture Control Objective
V1.5-Input and Output Architecture
V1.6-Cryptographic Architecture
V2-Authentication Control Objective
V2.2-General Authenticator Security
V3.3-Session Termination
V3.7-Defenses Against Session Management Exploits

tghosth:
Please see my question in the PR

ussjoin:
I responded in the PR, but duplicating here because it doesn't look like it got seen:

@tghosth I don't think it's a "problem" per se; it's your project, and you can of course govern it how you all choose. That said, if that's what you're doing, I would say that you should document it: https://github.com/OWASP/ASVS/blob/master/README.md doesn't mention that 4.0 is frozen, nor does https://github.com/OWASP/ASVS/blob/master/CONTRIBUTING.md, which are the two places I know to look from my work in other OSS projects.

If you're asking my personal opinion, I would propose accepting the changes in both 4.0 and 5.0, but not release a 4.0.4 until 5.0 is released (so you have a true "final" 4.0-series)---and at that time also removing the 4.0 files to an archival folder in the repo. This would let you not "focus" on 4.0 while allowing others who find other minor things like this---things that affect the readability of the standard, but not the content---continue to provide these tiny minor patches. Then when you release 5.0, you can declare, a la Knuth, that ["From that moment on, all 'bugs' will be permanent 'features.'"](https://www.tug.org/TUGboat/tb11-4/tb30knut.pdf)

tghosth:
Sorry, I did see but I had to think about it :S 

See here: https://github.com/OWASP/ASVS/pull/1257#issuecomment-1101554658

PR #1257, opened 2022-04-06
Several Typographical Errors in 4.0.3 and 5

ussjoin:
This Pull Request relates to issue #1256.

Fixing minor typographical errors in the Control Objectives of
several sections and subsections. No substantive changes and no
changes to individual controls.

[MODIFIED] V1-Architecture Control Objective
[MODIFIED] V1.5-Input and Output Architecture
[MODIFIED] V1.6-Cryptographic Architecture
[MODIFIED] V2-Authentication Control Objective
[MODIFIED] V2.2-General Authenticator Security
[MODIFIED] V3.3-Session Termination
[MODIFIED] V3.7-Defenses Against Session Management Exploits


tghosth:
Hi @ussjoin, thanks for taking the time. At this point, our inclination is to focus on changes to the 5.0 folder and leave 4.0 frozen at the 4.0.3 release. Do you think this is a problem?

ussjoin:
@tghosth I don't think it's a "problem" per se; it's your project, and you can of course govern it how you all choose. That said, if that's what you're doing, I would say that you should document it: <https://github.com/OWASP/ASVS/blob/master/README.md> doesn't mention that 4.0 is frozen, nor does <https://github.com/OWASP/ASVS/blob/master/CONTRIBUTING.md>, which are the two places I know to look from my work in other OSS projects.

If you're asking my personal opinion, I would propose accepting the changes in both 4.0 and 5.0, but not release a 4.0.4 until 5.0 is released (so you have a true "final" 4.0-series)---and at that time also removing the 4.0 files to an archival folder in the repo. This would let you not "focus" on 4.0 while allowing others who find other minor things like this---things that affect the readability of the standard, but not the content---continue to provide these tiny minor patches. Then when you release 5.0, you can declare, a la Knuth, that ["From that moment on, all 'bugs' will be permanent 'features.'"](https://www.tug.org/TUGboat/tb11-4/tb30knut.pdf)

tghosth:
Hi @ussjoin, thanks for your feedback. I completely agree with your point about [CONTRIBUTING.md](https://github.com/OWASP/ASVS/blob/master/CONTRIBUTING.md) and that came to mind immediately when I saw your PR. Do you think that the changes I propose in #1259 address this issue? I would welcome your feedback on that.

Regarding 4.0, as you can see I don't really want to try and maintain changes to that branch unless absolutely necessary so for now I would appreciate it if you could alter the PR to only modify the 5.0 folder.

tghosth (approved):

Issue #906, opened 2021-02-19
Control 2.2.1 > Too specific and tolerant


TheDauntless:
Concerning 2.2.1: https://github.com/OWASP/ASVS/blob/master/4.0/en/0x11-V2-Authentication.md#v22-general-authenticator-requirements

> Verify that anti-automation controls are effective at mitigating breached credential testing, brute force, and account lockout attacks. Such controls include blocking the most common breached passwords, soft lockouts, rate limiting, CAPTCHA, ever increasing delays between attempts, IP address restrictions, or risk-based restrictions such as location, first login on a device, recent attempts to unlock the account, or similar. **Verify that no more than 100 failed attempts per hour is possible on a single account.**

100 authentication requests before lockout already seems very generous, and would allow you to perform password spraying on a large list of popular passwords. In addition, the ASVS requirement says that it should be 'per hour', which insinuates that you can just try again after an hour, continuing your attack.

The 100 seems very arbitrary, but appears to come from the NIST guide: 
https://pages.nist.gov/800-63-3/sp800-63b.html#throttle

However, the NIST guide does not say '100 per hour' (but rather 100 at most). 

> When required by the authenticator type descriptions in Section 5.1, the verifier SHALL implement controls to protect against online guessing attacks. Unless otherwise specified in the description of a given authenticator, the verifier SHALL limit consecutive failed authentication attempts on a single account to no more than 100.

The 2.2.1 control is the only one that mentions account lockout, so there are no other controls that would put more stringent requirements on account locks. For a financial application, 100 is definitely too much, especially per hour. I think either this control should be modified to the text below, or an additional control should be added that focusses on account lockout, and this control should be rewritten to focus on actual throttling against password spraying and not brute-forcing/account lockouts.

So my suggestion, though I'm not sure if you use this risk profile approach in other controls:

> Verify that anti-automation controls are effective at mitigating breached credential testing, brute force, and account lockout attacks. Such controls include blocking the most common breached passwords, soft lockouts, rate limiting, CAPTCHA, ever increasing delays between attempts, IP address restrictions, or risk-based restrictions such as location, first login on a device, recent attempts to unlock the account, or similar. **Verify that only a limited number of consecutive unsuccessful authentication attempts are allowed, according to the risk profile of the application**

The actual number would then of course depend on the application, though some (much lower) defaults could be nice.

[...]

PR #1269, opened 2022-04-22
Control 2.2.1

TheDauntless:
Proposal for #906 

I only included the 'per hour' and not the total amount, since that one does not have an obvious 'reset' and it would be confusing to a user if they are at '99 failed attempts' for total lifetime, and then suddenly their account is blocked after a single typo.

5 attempts per hour could trigger a response, and typically that means sending an email to the end user allowing them to still log in, which will prevent account lockout as well.


tghosth:
@elarlang please review and let me know what you think

5.0/en/0x11-V2-Authentication.md
```diff
@@ -57,7 +57,7 @@ NIST considers SMS as ["restricted" authenticator types](https://pages.nist.gov/
...
-| **2.2.1** | Verify that anti-automation controls are effective at mitigating breached credential testing, brute force, and account lockout attacks. Such controls include blocking the most common breached passwords, soft lockouts, rate limiting, CAPTCHA, ever increasing delays between attempts, IP address restrictions, or risk-based restrictions such as location, first login on a device, recent attempts to unlock the account, or similar. Verify that no more than 100 failed attempts per hour is possible on a single account. | ✓ | ✓ | ✓ | 307 | 5.2.2 / 5.1.1.2 / 5.1.4.2 / 5.1.5.2 |
+| **2.2.1** | Verify that anti-automation controls are effective at mitigating breached credential testing, brute force, and account lockout attacks. Such controls include blocking the most common breached passwords, soft lockouts, rate limiting, CAPTCHA, ever increasing delays between attempts, IP address restrictions, or risk-based restrictions such as location, first login on a device, recent attempts to unlock the account, or similar. More than 5 failed authentication attempts per hour should trigger some sort of reaction or alert. | ✓ | ✓ | ✓ | 307 | 5.2.2 / 5.1.1.2 / 5.1.4.2 / 5.1.5.2 |
```
tghosth, line 60: ```should trigger some sort of reaction or alert.```

Based on @elarlang's question in #906, should this be:
- per user?
- per application?
- per IP?

tghosth (changes requested):

5.0/en/0x11-V2-Authentication.md
jmanico: I think 5 failed attempts per user is the goal here.

TheDauntless:
Definitely per user, so I've added it to the PR.

jmanico:
Folks, I think we are ready for a PR

tghosth (approved):

PR #1321, opened 2022-07-26
Clarify 8.2.2 to resolve #1141

### LINKED ISSUE #1141: Questions on rigidity of token-based session storage, "sessionStorage for token-based sessions" guidance in 8.2.2
Hi all, I read through https://github.com/OWASP/ASVS/issues/843 and https://github.com/OWASP/ASVS/issues/1091 to see that the bleeding edge has some exceptions for token-based session storage.

From the change to [8.2.2](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x16-V8-Data-Protection.md#v82-client-side-data-protection) it appears (from my understanding) that two types of sessions are mentioned:
- **cookie-based sessions**: Sessions where an application server sets the cookie with a Set-Cookie header (HTTP-Only), and the domain of the cookie is the recipient of authorized API requests. There is no reason why client JS would need to access the session, which is the basis for the clarification in [8.2.2](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x16-V8-Data-Protection.md#v82-client-side-data-protection).
- **token-based sessions**: Tokens used by browser clients (delegated access token, for example) that are attached as Authorization headers to outbound CORS requests. Token storage for these sessions in the browser seems to be a necessary evil, and the guidance is to use sessionStorage.

My question is, the session distinction makes sense, but should sessionStorage be mentioned here at all?

Storage for token-based sessions comes up in the first place because of two persistence needs:
1. Session retention across page refreshes
2. Session retention across browser tabs of the same application

The sessionStorage exception only solves persistence need 1, but not persistence need 2. The only path forward for these types of applications is to initiate new token-based sessions per tab, which requires a new login/authorization per tab. This doesn't really make the client more _secure_ from a storage perspective, since XSS can still steal tokens from sessionStorage, and also adds a lot of complexity to the app for session management. It, in my opinion, opens up even more questions (and potentially insecure solutions) about how to securely manage token-based sessions across tabs, which doesn't seem to be covered.

Should there be specific guidance about which browser storage to use for token-based sessions? All browser storage is inherently insecure because of XSS, so what benefit does sessionStorage provide that it is worth being called out?

| Storage | Susceptible to XSS | Susceptible to Token Exfiltration | Meets Persistence Need 1 | Meets Persistence Need 2 | Notes |
| ---- | ---- | ---- | ---- | ---- | ---- |
| Web Workers | Yes | No | :white_check_mark: | :white_check_mark: | Accessible across tabs if SharedWorker is used. Potentially complex to set up. |
| In-memory Closures | Yes | No (?) | - | - |  Only available within current tab. |
| Session Storage | Yes | Yes | :white_check_mark: | - | Only available within current tab. |
| Cookie - Samesite=Strict, Secure, _not_ HTTP-Only (essentially "stricter" local storage) | Yes | Yes | :white_check_mark: | :white_check_mark: | Expires when token expires if expiration is set (or additionally when browser is closed if set as a session cookie). If used purely for storage, does not open itself up for CSRF, because outbound requests require attached Auth Headers. |
| Local Storage | Yes | Yes | :white_check_mark: | :white_check_mark: | Accessible within domain. No expiration, needs to be manually cleared by the client, but [8.2.3](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x16-V8-Data-Protection.md#v82-client-side-data-protection) covers this. |

Should the distinction instead be something along the lines of:
> Verify that data stored in browser storage (such as localStorage, sessionStorage, IndexedDB, or cookies) does not contain sensitive data, with the exception of cookie-based session tokens and token-based session tokens, with the former stored only in cookies, following [V3.4](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x12-V3-Session-management.md#v34-cookie-based-session-management).

Very curious on thoughts here. Also, let me know if there is another discussion somewhere discussing this.

tghosth:
This Pull Request relates to issue #1141

### DIFF
--- a/5.0/en/0x16-V8-Data-Protection.md
+++ b/5.0/en/0x16-V8-Data-Protection.md
@@ -28,7 +28,7 @@ Ensure that a verified application satisfies the following high level data prote
 | # | Description | L1 | L2 | L3 | CWE |
 | :---: | :--- | :---: | :---: | :---: | :---: |
 | **8.2.1** | Verify the application sets sufficient anti-caching headers so that sensitive data is not cached in modern browsers. | ✓ | ✓ | ✓ | 525 |
-| **8.2.2** | [MODIFIED] Verify that data stored in browser storage (such as localStorage, sessionStorage, IndexedDB, or cookies) does not contain sensitive data, with the exception of session identifiers which should be stored in either cookies or sessionStorage. | ✓ | ✓ | ✓ | 922 |
+| **8.2.2** | [MODIFIED] Verify that data stored in browser storage (such as localStorage, sessionStorage, IndexedDB, or cookies) does not contain sensitive data, with the exception of session tokens which should be stored in either cookies or sessionStorage. | ✓ | ✓ | ✓ | 922 |
 | **8.2.3** | [MODIFIED] Verify that authenticated data is cleared from client storage, such as the browser DOM, after the client or session is terminated. The "Clear-Site-Data header" may be able to help with this but the client-side should also be able to clear up if the server connection is lost. | ✓ | ✓ | ✓ | 922 |
 
 ## V8.3 Sensitive Private Data
--- a/5.0/en/0x16-V8-Data-Protection.md
+++ b/5.0/en/0x16-V8-Data-Protection.md
@@ -28,7 +28,7 @@ Ensure that a verified application satisfies the following high level data prote
 | # | Description | L1 | L2 | L3 | CWE |
 | :---: | :--- | :---: | :---: | :---: | :---: |
 | **8.2.1** | Verify the application sets sufficient anti-caching headers so that sensitive data is not cached in modern browsers. | ✓ | ✓ | ✓ | 525 |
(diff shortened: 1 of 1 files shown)


elarlang:
my comment in issue: https://github.com/OWASP/ASVS/issues/1141#issuecomment-1208034579

elarlang (commented):
session identifiers > session tokens

tghosth:
@elarlang you think we should be saying session tokens instead of session identifers?

elarlang:
> @elarlang you think we should be saying session tokens instead of session identifiers?

yes, as rest of the ASVS call it "session token" everywhere (3.4.1, 3.4.2, 3.4.3, 3.4.4, 3.4.5, 3.4.6, 3.5.2, 3.5.3, 7.1.1, 7.2.1, 8.2.3)

tghosth:
Fair point @elarlang, is that ok now? :)

elarlang (approved):

PR #1330, opened 2022-08-11
Merge 3.1.1 and 13.1.3 into 8.3.1 to resolve #1313

### LINKED ISSUE #1313: 8.3.1 and 13.1.3 about data in URLs/query string seem to be duplicated
Can anyone see why we should not merge these two requirements?

| # | Description |
| :---: | :--- |
| [8.3.1](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x16-V8-Data-Protection.md#v83-sensitive-private-data) | Verify that sensitive data is sent to the server in the HTTP message body or headers, and that query string parameters from any HTTP verb do not contain sensitive data. |
| [13.1.3](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x21-V13-API.md#v131-generic-web-service-security) | Verify API URLs do not expose sensitive information, such as the API key, session tokens etc. |

tghosth:
This Pull Request relates to issue #1313


Sjord:
> Verify that sensitive data is only sent to the server in the HTTP message body or headers

Should we consider websockets here? They are fine to send sensitive data, but I would not consider them message body or headers.

> and that the URL/query string do

I don't like the slash in "URL/query string" part. I think it makes it a little bit unclear, both in meaning and whether this is singular or plural

> not contain sensitive information, such as an API key, session token, etc. 

"such as" already indicates that this list is not exhaustive. Is "etc." still needed?

I would write it like so:

> Verify that sensitive data is only sent to the server in the HTTP message body or headers and that the URL and query string do not contain sensitive information, such as an API key or session token. 

But this is a little bit nitpicking and if this was merged as is I would also be fine with it.

elarlang (approved):

elarlang:
@tghosth - I approved and you can resolve conflicts now :)

Sjord - please move your remarks to related issue, we can do one more finetuning round if needed.

PR #1356, opened 2022-08-30
German translation from @JoergBruenner

tghosth:


tghosth:
I have opened a different PR with some issues addressed in #1357

Issue #1381, opened 2022-09-28
8.1.3 seems like nonsense


jmanico:
8.1.3 makes absolutely no sense

From: https://github.com/OWASP/ASVS/blob/master/5.0/en/0x16-V8-Data-Protection.md

8.1.3 | [GRAMMAR] Verify the application minimizes the number of parameters in a request, such as hidden fields, Ajax variables, cookies, and header values. |   | ✓ | ✓ | 233
-- | -- | -- | -- | -- | --

elarlang:
The sense for example - instead of keeping already submitted data on the service side, some app carries entire package in hidden fields / in some JSON object etc.

It is problematic also with some restful stuff, where entire datasets are sent back and forward when only few items are needed.

jmanico:
> The sense for example - instead of keeping already submitted data on the service side, some app carries entire package in hidden fields / in some JSON object etc.
> 
> It is problematic also with some restful stuff, where entire datasets are sent back and forward when only few items are needed.

Server vs client-side data is a different issue covered elsewhere. I do not think the field number really matters. This, IMO, needs to be clarified or just removed. We have bigger issues to focus on.

elarlang:
From the "data protection" category perspective this requirement is at least in wrong place. We can minimize unnecessary bandwidth from client to server with that, but it's not a data protection issue.

jmanico:
And having many parameters vs JSON is actually SAFER since you do not need to utilize a server-side JSON parser.

I again assert this requirement is nonsense. Many apps require lots of parameters for good reason.

I do see EXTRA parameters the app does not need as a good intrusion detection point. But many parameters is just how some apps work, and that does not add a security risk in an of itself.

jsulinski:
I agree, this requirement is too vague to be useful.

tghosth:
Do we have any idea what the original rationale was here? Can we see from blame or previous issues?

jsulinski:
Doesn't seem to exist in ASVS [1.0](https://github.com/OWASP/ASVS/blob/master/1.0/asvs-webapp-release-2009-en.pdf).
Exists in ASVS [2.0](https://github.com/OWASP/ASVS/blob/master/2.0/OWASPApplicationSecurityVerificationStandard3.0.pdf) as 9.7.
Exists in ASVS [3.0.1](https://github.com/OWASP/ASVS/blob/master/3.0.1/OWASP%20Application%20Security%20Verification%20Standard%203.0.1.pdf) as 9.7.
[...]

PR #1387, opened 2022-09-30
remove 8.1.3 (#1381)

elarlang:
This Pull Request relates to issue #1381


danielcuthbert (approved):
This is a legacy requirement and I'm not even sure how it made it in. It's vague and just doesn't add any value. Agree to remove

Issue #1385, opened 2022-09-29
5.3.5 is dead wrong and this may impact the fate of the planet


jmanico:
PHP dropped its manual escaping functions years ago because - they do not work.
This requirement needs to go away!

https://www.php.net/manual/en/function.mysql-real-escape-string.php

elarlang:
Requirement itself, PR for removing it on the way
| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **5.3.5** | Verify that where parameterized or safer mechanisms are not present, context-specific output encoding is used to protect against injection attacks, such as the use of SQL escaping to protect against SQL injection. ([C3, C4](https://owasp.org/www-project-proactive-controls/#div-numbering)) | ✓ | ✓ | ✓ | 89 |

PR #1388, opened 2022-09-30
removed 5.3.5 (#1385)

elarlang:
This Pull Request relates to issue #1385


danielcuthbert (approved):
Agree with the deletion of this

Issue #1382, opened 2022-09-28
Discuss MTLS for internal services


jmanico:
I think we need to get more aggressive on suggesting the use mTLS intra-service for level 3.

Please note mTLS - Mutual TLS.
Somewhere here.
https://github.com/OWASP/ASVS/blob/master/5.0/en/0x17-V9-Communications.md in 9.3

csfreak92:
+1 on this, I totally agree we should suggest mTLS. We usually see this in security and architecture reviews that most setups do not have mTLS for intra-service communications.

tghosth:
@csfreak92 do you fancy drafting some wording?

csfreak92:
Hi @tghosth, I can give it a shot. Please assign to me this issue. Also, are we just suggesting this for Level 3? Wouldn't it make more sense for Level 2 requirement as well?

tghosth:
Lets start with L3 as it is not super trivial. Assigned to you

PR #1408, opened 2022-10-27
Update 0x17-V9-Communications.md

csfreak92:
Adding 9.3.3 for mTLS requirement. CWE 295 is the closest I can match this requirement with. 

This Pull Request relates to issue #1382.

### DIFF
--- a/5.0/en/0x17-V9-Communications.md
+++ b/5.0/en/0x17-V9-Communications.md
@@ -48,7 +48,7 @@ HTTP traffic between internal facing services should also be encrypted, idealy u
 | :---: | :--- | :---: | :---: | :---: | :---: |
 | **9.3.1** | [ADDED] Verify that TLS or another appropriate transport encryption mechanism used for all connectivity between internal, HTTP-based services, and does not fall back to insecure or unencrypted communications. | | ✓ | ✓ | 319 |
 | **9.3.2** | [MODIFIED, MOVED FROM 9.2.1] Verify that TLS connections between internal services use trusted certificates. Where internally generated or self-signed certificates are used, the consuming service must be configured to only trust specific internal CAs and specific self-signed certificates. All others should be rejected. | | ✓ | ✓ | 295 |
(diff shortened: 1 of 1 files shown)


tghosth (changes requested):
1) Do we want to specify that we are talking about internally within the system as opposed to ok every external connection as well.
2) I don't want to use the expression zero trust so I would consider removing that last sentence.

csfreak92:
@tghosth, clarifications on your questions: 

1. I think intra-service communication should be doing mTLS mainly as more applications are in cloud-hosted environments these days, it would improve the security posture of those systems. Regarding have external service-to-service connections, I think you have a point too. So shall I change it to accommodate both internal and external service-to-service connections instead? If that's the case, then it should fall in 9.2.x section then. 
2. I'm just curious as to why you don't like using the zero trust expression in this context? I might be missing something that I do not understand.

3m29xp2 (approved):
> `gh pr checkout 

> 1408`

tghosth:
Hi @csfreak92 :)

> I think intra-service communication should be doing mTLS mainly as more applications are in cloud-hosted environments these days, it would improve the security posture of those systems. Regarding have external service-to-service connections, I think you have a point too. So shall I change it to accommodate both internal and external service-to-service connections instead? If that's the case, then it should fall in 9.2.x section then.

To be honest, my point was that we should focus on internal because requiring mTLS for external would be a lot more problematic as it would require all external/3rd party services to support it which is unrealistic.

> I'm just curious as to why you don't like using the zero trust expression in this context? I might be missing something that I do not understand.

It is just a bit of a buzzword that does really mean anything in this context. If you want to give a reason for doing this, we should be more specific.

csfreak92:
@tghosth, I am inclined to agree that we stick with requiring/promoting "intra-service" mTLS communications for now and then re-evaluate that requirement in the future to add in case we decide to have one for external services communicating with each other since that is quite a more problematic task to deal with implementation. 

Regarding zero trust expression, gotcha I will remove that then.

csfreak92:
Updated my PR and commit @tghosth, @elarlang. For your review please.

tghosth (approved):

Issue #1409, opened 2022-10-31
Incorrect CWE allocation


vdbaan:
Within [chapter 8](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x16-V8-Data-Protection.md)

You have the two following requirements:

8.3.3 | L1 | Data Protection | Sensitive Private Data | Verify that users are provided clear language regarding collection and use of supplied personal information and that users have provided opt-in consent for the use of that data before it is used in any way.
-- | -- | -- | -- | --
8.3.8 | L2,L3 | Data Protection | Sensitive Private Data | Verify that sensitive personal information is subject to data retention classification, such that old or out of date data is deleted automatically, on a schedule, or as the situation requires.

This is all mapped to CWE 285 which has the following text and Title.

CWE-285: Improper Authorization: The software does not perform or incorrectly performs an authorization check when an actor attempts to access a resource or perform an action.

The CWE does not match the language regarding data protection, content, or retention.

tghosth:
@elarlang is there no CWE that would match in this case?

tghosth:
@vdbaan do you think [CWE-359 - "Exposure of Private Personal Information to an Unauthorized Actor"](https://cwe.mitre.org/data/definitions/359.html) is close enough?

elarlang:
CWE-359 is more like pointing to authorization problem, but this is not (directly) the case here.

vdbaan:
The CWE does not relate to the requirements.
The requirements to me are about customer control over their own data (think GDPR).
The CWE reads to me about exposure of any users sensitive data, slightly different angle.

PR #1414, opened 2022-11-05
remove incorrect CWE values

elarlang:
This Pull Request relates to issue #1409


tghosth (changes requested):
@elarlang is there no CWE that would match in this case?

elarlang:
I have not done any research on that. Just removing the incorrect ones.

tghosth:
OK so let's see if the discussion in #1409 gets anywhere...

elarlang:
Point of view is still the same, to just remove incorrect CWE values.

tghosth:
Do you really think we cannot find a relevant CWE?

elarlang:
as CWE is for weakness identifier, but those requirements coming from policy without being really weaknesses, the answer is yes.

tghosth (approved):

Issue #1469, opened 2022-12-22
new requirement for Formula Injection


RafaelGreen1:
Hi,
In [V5.3 Output Encoding and Injection Prevention](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x13-V5-Validation-Sanitization-Encoding.md#v53-output-encoding-and-injection-prevention) I would have expected to see something about formula injection.
For example, when a server of summer camps website genrates xlsx file with students' attendace data, last name of one the students could be something like "=HYPERLINK(url, student's last name)" or "=cmd|' /C...".
There are explanations about this attack here: [CSV_Injection](https://owasp.org/www-community/attacks/CSV_Injection).
Does it appear elsewhere in the standard?
Thanks,
Rafael,
Edit: Thanks to Haim Mitrany who informed me about this vulnerability.

elarlang:
Nope, we don't have separate requirement for that. But we should have one :)

elarlang:
Also note, for the requirement we should avoid the name CSV injection as it is misleading - is not only related to CSV, it works in xls, xlsx, odf, Google Spreadsheet etc.

Defense method is escaping.

tghosth:
2 linked problems that probably need to be solved together because the mitigation for formula injection requires the mitigation for CSV injection to also be applied in order to be effective.

tghosth:
Hi @RafaelGreen1, are you able to suggest a requirement wording for this new requirement?

@set-reminder 3 weeks @tghosth to revisit

RafaelGreen1:
Hi @tghosth , I created this pull request: https://github.com/OWASP/ASVS/pull/1482
(I tried to keep the requirement short, but I added a link in the References).
It's my first pull request here, hope it's ok.
Thanks,
Rafael,

RafaelGreen1:
@elarlang, following your comment in the pull request, I submitted a commit to fix the reference label in 1.5.4.
I understand from your comment that we need to finish the discussion here before doing the pull request. But I saw a 'Changes requested' label so I created a fix commit (If you want, I can cancel the pull request and make a new one after we finish the discussion).

For your convenience, I put here the suggested wording:

[...]

PR #1482, opened 2022-12-28
Add new requirement for Formula Injection

RafaelGreen1:
This Pull Request relates to issue #1469


elarlang (dismissed):

elarlang (changes requested):
please do not change numbering OR there is need to fix the reference label in 1.5.4 requirement. We going to renumber requirements before v5.0 release anyway (to fill all the caps)

With proposal we think requirement text proposal in issue :)

elarlang:
I close this one. We will have new one when agreement and go-ahead in the related issue.

Issue #1255, opened 2022-04-05
8.3.6 on memory wiping


jmanico:
8.3.6 | Verify that sensitive information contained in memory is overwritten as soon as it is no longer required to mitigate memory dumping attacks, using zeroes or random data.
-- | --

Several of my students believe that this is not a reasonable requirement that actually helps security or is a real vulnerability.

I suggest we remove it.

Sjord:
I agree. I looked into this earlier:

* [Clearing secrets from memory](http://www.sjoerdlangkemper.nl/2016/05/22/should-passwords-be-cleared-from-memory/)
* [Overwriting memory in Python](http://www.sjoerdlangkemper.nl/2016/06/09/clearing-memory-in-python/)

It primarily reduces impact in the case of a buffer overread (i.e. [heartbleed](https://en.wikipedia.org/wiki/Heartbleed)), but it is often impossible to reliably clear memory, and is not worth the hassle in my opinion.

tghosth:
@set-reminder 3 weeks @tghosth to look at this

tghosth:
So I think that 8.3.8 is quite general and doesn't exactly what fit the intention here. 

I agree that 8.3.6 should not be a sweeping requirement but maybe we could make it more specific.

| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---:| :---: | :---: |
| **8.3.6** | Where a memory safe language is not being used or where the language specifically supports this ability, verify that sensitive information contained in memory is overwritten as soon as it is no longer required using zeroes or random data, to mitigate memory dumping attacks. | | ✓ | ✓ | 226 |

@Sjord @jmanico what do you think about that?

Sjord:
I don't think it makes sense to have different security requirements for different programming languages. The property of memory-safety does not define whether memory can be reliably overwritten. It's hard to reliably overwrite memory in any language, as far as I know.

Do you have a concrete example of which function you should call in which programming language to clear memory?

tghosth:
@Sjord I accept your point, I thought I saw something in rust but I take the point that there are lots of reasons why this wouldn't work. I am going to try and solicit some wider feedback.. :)

iloving:
So this discussion has two components, and they need to be addressed separately:
1. Is this a real risk? The short answer is yes.  If you are able to access processes memory in any way, you can get at this data.  Possible modes of access are, off the top of my head:
 - As already mentioned, direct exploits such as buffer overreads
[...]

PR #1497, opened 2023-01-03
Remove 8.3.6 as impractical

tghosth:
This Pull Request relates to issue #1255

@elarlang please review


elarlang (changes requested):
unnecessary `.` after label and need to decide label first

elarlang (approved):

PR #1499, opened 2023-01-03
Resolve #1394 by clarifying 2.2.2

### LINKED ISSUE #1394: [Question] Does V2.2.2 allow magic links?
I have a question regarding requirement [2.2.2](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x11-V2-Authentication.md#v22-general-authenticator-security). Here it says: 

> Verify that the use of weak authenticators (such as SMS and email) is limited to secondary verification and transaction approval [...]

Does this mean that authentication via [magic link](https://auth0.com/docs/authenticate/passwordless/authentication-methods/email-magic-link) is prohibited?

If so, why is it prohibited?
Most applications allow resetting a password via email. So if an attacker gains access to an email account, they can simply reset the password. 
Therefore, SFA via magic link doesn't seem more vulnerable to me than SFA via password.

tghosth:
This is a work in progress for now


tghosth:
Ok, there were no responses to my points in https://github.com/OWASP/ASVS/issues/1394#issuecomment-1478521746 so let's just get this merged @elarlang

elarlang (approved):
approve, but "email is not used" is still not clear.

Issue #1510, opened 2023-01-04
2.4.* - remove "as strong as verification server performance allow" part from requirements


elarlang:
As an example with `bcrypt`, in ASVS v4.0.3 we had:
_V2.4.4 Verify that if bcrypt is used, the work factor SHOULD be as large as verification server performance will allow, with a minimum of 10._

I suggested to remove "_as large as verification server performance will allow, with a_" part (https://github.com/OWASP/ASVS/issues/1159#issuecomment-1175297625)

and we have now:
_Verify that if bcrypt is used, the work factor is a minimum of 10 and password size is limited to 72-bytes due to bcrypt's input limit._

I think we should get rid of this part from other similar requirements as well, as it's not possible to verify that in practice:

| # | Description | L1 | L2 | L3 | CWE | [NIST &sect;](https://pages.nist.gov/800-63-3/sp800-63b.html) |
| :---: | :--- | :---: | :---: | :---: | :---: | :---: |
| **2.4.3** | [MODIFIED] Verify that if PBKDF2 is used, the iteration count should **be as large as verification server performance will allow, with** a minimum of 720,000 iterations with PBKDF2-HMAC-SHA1, a minimum of 310,000 iterations using PBKDF2-HMAC-SHA256, or with a minimum of 120,000 iterations with PBKDF2-HMAC-SHA512. ([C6](https://owasp.org/www-project-proactive-controls/#div-numbering)) | | ✓ | ✓ | 916 | 5.1.1.2 |
| **2.4.4** | [MODIFIED] Verify that if bcrypt is used, the work factor is a minimum of 10 and password size is limited to 72-bytes due to bcrypt's input limit. ([C6](https://owasp.org/www-project-proactive-controls/#div-numbering)) | | ✓ | ✓ | 916 | 5.1.1.2 |
| **2.4.5** | [DELETED] | | | | | |
| **2.4.6** | [ADDED] Verify that if argon2id is used, the configuration should **be as strong as verification server performance will allow, with** a minimum configuration of 15 MiB of memory, an iteration count of 2, and 1 degree of parallelism. ([C6](https://owasp.org/www-project-proactive-controls/#div-numbering)) | | ✓ | ✓ | 916 | 5.1.1.2 |
[...]

PR #1532, opened 2023-01-26
removed "as strong as verification server performance allow" part from requirements and modified the en docs as per #1510

suyash5053:
This PR removed "as strong as verification server performance allow" part from requirements and modified the en docs as per #1510  .


elarlang:
. o O ( what is going on here? )

We have certain procedure and rules for making changes, the issue is assigned to Josh and he will handle it when he reach to this issue with task-list.

Thank you for effort - we wait PR's mostly from users if they proposed something themselves.

suyash5053:
@elarlang Really Sorry for the inconvenience, I caused, all this happened due to the excitement of contribution. Sorry for the informal behaviour.

Issue #1567, opened 2023-03-07
2.4.3 PBKDF2 iterations value should be changed from 100,000 to 600,000 in alignment with OWASP Password Storage Cheat Sheet


markonweb:

markonweb:
Proposal to change 2.4.3 (line 98) PBKDF2 iterations value from 100,000 to 600,000 in alignment with the OWASP Password Storage Cheat Sheet https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

elarlang:
What is the argumentation behind the value "600,000"? Other than it is in the cheat sheet.

markonweb:
Good question. The cheat sheet recommendation points to https://tobtu.com/minimum-password-settings/ , a page on a site owned by Steve "sc00bz" Thomas. Also on the site, is a link his Blackhat 2022 presentation https://tobtu.com/files/blackhat2022.pdf  
Sides 18-20 are about PBKDF2 including a minimum strength recommendation of "Such that an attacker gets <10 KH/s/GPU" with a footnote (17) that links to an ArsTechnica article about the 2015 LastPass breach https://arstechnica.com/?p=685505 ... I haven't seen what led to the 600,000 iteration recommendation.
fwiw, the cheat sheet iteration recommendation carried into the wikipedia entry for pbkdf2 as an "OWASP recommendation" https://en.wikipedia.org/wiki/PBKDF2
RFC 2898 (Sept 2000) isn't prescriptive except that it recommends a min of 1000 iterations. https://www.rfc-editor.org/rfc/rfc2898#section-4
RFC 8018 (Jan 2017) obseletes RFC 2898 but restates the same numbers and section 4.2 references NIST SP 800-132 as the source. Min of 1000 iterations https://www.rfc-editor.org/info/rfc8018
CWE 916 isn't prescriptive in terms of iterations https://cwe.mitre.org/data/definitions/916.html
NIST SP 800-132 (Dec 2010) isn't prescriptive either, but also recommends a min of 1000 iterations in section 5.2. https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-132.pdf
Note - NIST just closed a public comment period on SP 800-132 (so there will likely be changes) https://content.govdelivery.com/accounts/USNIST/bulletins/3414e52
[...]

Issue #1508, opened 2023-01-04
Update for PBKDF2 settings


jmanico:
PBKDF2 iterations were reset a bit by the main mathematician from HashCat who drives international standards (IETF) on this topic.

Looking into the past commits to the cheat sheet it seems the current ones were sourced from:

https://twitter.com/sc00bzt/status/1322164349738262528

These recommendations were updated again this December:

Parallel PBKDF2
PPBKDF2-SHA512: cost 2 (Based on RTX 4090 as "1.5 GPUs")
PPBKDF2-SHA256: cost 5 (Based on RTX 4090 as "1.5 GPUs")
PPBKDF2-SHA1: cost 10 (Based on RTX 4090 as "1.5 GPUs")

PBKDF2 (worst)
PBKDF2-HMAC-SHA512: 210,000 iterations (Based on RTX 4090 as "1.5 GPUs")
PBKDF2-HMAC-SHA256: 600,000 iterations (Based on RTX 4090 as "1.5 GPUs")
PBKDF2-HMAC-SHA1: 1,300,000 iterations (Based on RTX 4090 as "1.5 GPUs")

https://infosec.exchange/@sc00bz/109599362314030488

See also: https://tobtu.com/minimum-password-settings/

PR #1568, opened 2023-03-07

markonweb:
Propose changing iterations value in **2.4.3** (line 98) from 100,000 to 600,000 in alignment with https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

This Pull Request relates to issue #1567


elarlang:
I close this PR at the moment, I think we need some wider change like proposed in #1508.

In genera, let's handle PR if there is agreement in related issue.

Issue #1188, opened 2022-02-02
2.5.4 split it up


jmanico:
The process of verifying a shared password for admin accounts - and the process of verifying default admin accounts - is very different and henceforth we should break this up like the Beatles.

tghosth:
Just took a trip down memory lane to try and figure out why it was like that in the first place and why it has that NIST identifier.

Shared accounts seems to have been added to an existing requirement about spoofing here:
https://github.com/OWASP/ASVS/commit/c0f9c295ea57ecfba848cfbdca850034825117f1

It seems to have become shared or default here:
https://github.com/OWASP/ASVS/commit/6298456976626a1853f296f5c892f78c3b5379ec

I still don't know why it has that NIST reference.

CWE is very generic as well.

AND the section doesn't make sense to me anyway...

I actually think this is a good candidate for section V1 as it is a more pervasive thing across an application.

Currently we have:

[2.5.4](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x11-V2-Authentication.md#v25-credential-recovery):

> Verify shared or default accounts are not present (e.g. "root", "admin", or "sa").

I would propose adding the default aspect to [1.2](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x10-V1-Architecture.md#v12-authentication-architecture):

1.2.5 [MOVED AND SPLIT FROM 2.5.4] Verify that default user accounts are not present in the application or are disabled.

I am not actually sure if "shared accounts" should be included in the standard as it is not really related to how an application is built.

Thoughts?

flo-blg:
> I am not actually sure if "shared accounts" should be included in the standard as it is not really related to how an application is built.

Hi guys, I totally agree with this. Maybe I misunderstand this requirement but if I get it right, how are we supposed to prevent users from being able to create an account using a generic email address such as "contact@mycompany.com" or "team@mycompany.com"?

elarlang:
Another issue for the same requirement: https://github.com/OWASP/ASVS/issues/1395

tghosth:
@set-reminder 5 weeks look at this

tghosth:
Opened PR #1578

elarlang:
Re-open for level validation.

@tghosth - was level change from Level 1 to Level 2 intended or it happened as a mistake?

https://github.com/OWASP/ASVS/pull/1578/files/82e25bc952c25be950a5542d578dc529a83383be

tghosth:
[...]

PR #1578, opened 2023-03-21
Move default accounts to 1.2.5 and modify

tghosth:
This Pull Request relates to issue #1188


tghosth:
@elarlang  @atluxity @mgargiullo @flo-blg What do you think about this?

flo-blg:
> @elarlang @atluxity @mgargiullo @flo-blg What do you think about this?

It looks good to me, thanks!

elarlang (changes requested):
Requirement text is ok.

But I don't think it belongs to V1 - what is the reasoning to move it there? It should stay in V2 as it is impementation check and authentication requirement.

Issue #1545, opened 2023-02-17
1.6.4 grammar police strikes again


jmanico:
1.6.4 is still using bad grammar

1.6.4 | [GRAMMAR] Verify that the architecture treats client-side secrets - such as symmetric keys, passwords, or API tokens - as insecure and never uses them to protect or access sensitive data.
-- | --

possible fix (drop the dashes please):

1.6.4 | [GRAMMAR] Verify that the architecture treats client-side secrets such as symmetric keys, passwords, or API tokens as insecure and never uses them to protect or access sensitive data.
-- | --

tghosth:
@jmanico what do you think about:

1.6.4 | [GRAMMAR] Verify that the architecture treats client-side secrets, (such as symmetric keys, passwords, or API tokens,) as insecure and never uses them to protect or access sensitive data.
-- | --

elarlang:
Can you give an example, where this mistake can happen or how it is important?

jmanico:
I am not debating the validity of the requirement, I am concerned that this existing requirement has bad grammar. That's it! If you want to discuss the details I suggest you open a new issue.

tghosth:
Opened a PR to fix grammar

PR #1635, opened 2023-05-23
Fix grammar of 1.6.4

tghosth:
Reference #1545


elarlang (commented):
`, (such as symmetric keys, passwords, or API tokens,)` - this is really confusing syntax (usage of `(),`), even if it's grammatically correct

tghosth:
What would you prefer? I am willing to be less correct but clearer :)

elarlang:
Verify that the architecture treats client-side secrets (such as symmetric keys, passwords, or API tokens) as insecure and never uses them to protect or access sensitive data.

tghosth:
OK so should be ok now @elarlang

elarlang (approved):

elarlang:
@jmanico maybe want to verify that, as he opened the issue

Issue #1417, opened 2022-11-09
New Requirement on Sessions


jmanico:
A requirement as follows may be prudent.

_Make sure all active sessions or tokens are revoked when a user account is terminated or deleted (such as an employee leaving the company)._

elarlang:
It take content from other similar requirement #1418 :
_If you are using the access control mechanism of JWT's then be certain to revoke current active tokens when admins change the entitlements/roles of a user._

In my opinion it makes sense to combine them to one requirement or at least handle together.

tghosth:
I will take a stab at this

@set-reminder 5 weeks look at this

tghosth:
I think it is hopefully clear that section 3 requires the ability to logout and invalidate the session token, even if this is a JWT.

I don't however agree that these should be combined as I think they handle slightly different cases. The issue from #1418 is specifically relevant when using the contents of JWTs and other stateless tokens to make access control decisions. I was also wondering this requirement should be in chapter V4 because it is about access control but I think that given the key action here is on the session identifier, section 3 is probably more appropriate.

As such, I would suggest the following:

| # | Description | L1 | L2 | L3 | CWE | [NIST &sect;](https://pages.nist.gov/800-63-3/sp800-63b.html) |
| :---: | :--- | :---: | :---: | :---: | :---: | :---: |
| **[3.3.6](https://appsecg.host/3.3)** | [ADDED] Verify that all active sessions are revoked when a user account is disabled or deleted (such as an employee leaving the company). | ✓ | ✓ | ✓ | 613 | |
| **[3.5.7](https://appsecg.host/3.5)** | [ADDED] Verify that all active stateless tokens, which are being relied upon for access control decisions, are revoked when admins change the entitlements/roles of the user. | ✓ | ✓ | ✓ | 613 | |

@jmanico @elarlang What do you think?

jmanico:
> I think it is hopefully clear that section 3 requires the ability to logout and invalidate the session token, even if this is a JWT.
> 
> I don't however agree that these should be combined as I think they handle slightly different cases. The issue from #1418 is specifically relevant when using the contents of JWTs and other stateless tokens to make access control decisions. I was also wondering this requirement should be in chapter V4 because it is about access control but I think that given the key action here is on the session identifier, section 3 is probably more appropriate.
> 
> As such, I would suggest the following:
> 
> #	Description	L1	L2	L3	CWE	[NIST §](https://pages.nist.gov/800-63-3/sp800-63b.html)
[...]

Issue #1491, opened 2023-01-01
Using `/` instead of "or" in requirement text


elarlang:
Using `/` instead of "or" in requirement text can be confusing, as it has special syntax and meaning in some cases.

Valid uses of `/`:

* V13.2.5 Verify that REST services explicitly check the incoming Content-Type to be the expected one, such as application/xml or application/json.
* V14.1.1 Verify that the application build and deployment processes are performed in a secure and repeatable way, such as CI / CD automation, automated configuration management, and automated deployment scripts.
* V14.4.1 Verify that every HTTP response contains a Content-Type header. Also specify a safe character set (e.g., UTF-8, ISO-8859-1) if the content types are text/*, /+xml and application/xml. Content must match with the provided Content-Type header.
* V14.6.3 [ADDED] Verify that a full CRLF (\r\n) sequence is neutralized inside a HTTP/2 header.

Questionable:

* V2.10.2 [GRAMMAR] Verify that if passwords are required for service authentication, the service account used is not a default credential (e.g. root/root or admin/admin are default in some services during installation).
* V2.1.14 [ADDED, SPLIT FROM 2.1.7, LEVEL L1 > L3] Verify that passwords submitted during account registration or password changes are checked against a set of breached username/password pairs.

2.10.2 maybe should use `:`

Instead of "or" or "and":

* V1.4.5 Verify that attribute or feature-based access control is used whereby the code checks the user's authorization for a feature/data item rather than just their role. Permissions should still be allocated using roles.
* V1.14.5 Verify that application deployments adequately sandbox, containerize and/or isolate at the network level to delay and deter attackers from attacking other applications, especially when they are performing sensitive or dangerous actions such as deserialization.
* V2.4.7 [ADDED] Verify that if scrypt is used, the configuration SHOULD be as strong as verification server performance will allow, with a minimum CPU/memory cost parameter of (2^16), a minimum block size of 8 (1024 bytes), and a parallelization parameter of 1.
* V3.3.3 [LEVEL L2 > L1] Verify that the application gives the option to terminate all other active sessions after a successful password change (including change via password reset/recovery), and that this is effective across the application, federated login (if present), and any relying parties.
* V3.5.7 [ADDED] Verify that other, security-sensitive attributes of a stateless token are being verified. For example, in a JWT this may be the issuer, subject, and/or audience.
[...]

PR #1641, opened 2023-05-31
Add new session revocation requirements as per #1417

tghosth:
This Pull Request relates to issue #1417


elarlang (commented):
tiny style thing - can we avoid "/" instead of or? ("entitlements/roles")

See https://github.com/OWASP/ASVS/issues/1491

elarlang (approved):

PR #1669, opened 2023-06-15
Add 5.2.9 for sanitizing regexes to resolve #1600

### LINKED ISSUE #1600: Incomplete Regular Expression for Hostnames: Unescaped Dot Character
There is a vulnerability in some applications where the dot character is not properly escaped in regular expressions designed to validate hostnames. This oversight can lead to bypassing security checks, such as those for preventing attacks like request forgeries and malicious redirections. Incomplete checks may also cause undesirable behaviors when they accidentally succeed. 
If an attacker successfully exploits this vulnerability, they may perform unauthorized actions like forging requests or redirecting users to malicious websites. This can lead to a range of security issues, including data breaches, compromised application functionality, and exposing users to additional risks.

Suggested ASVS check:
_Verify that the application properly escapes special characters, such as the dot character, in regular expressions used for hostname validation to prevent security threats like request forgeries and malicious redirections._

Sample vulnerable code:
```
app.get('/some/path', function(req, res) {
    let url = req.param('url'),
        host = urlLib.parse(url).host;
    // BAD: the host of `url` may be controlled by an attacker
    let regex = /^((www|beta).)?example.com/;
    if (host.match(regex)) {
        res.redirect(url);
    }
});
```

tghosth:
This Pull Request relates to issue #1600


elarlang (commented):
"CWE-94: Improper Control of Generation of Code ('Code Injection')" does not feel correct here. I think if we don't have good candidate, we leave it empty (as there is plan to drop i tanyway)

tghosth:
@elarlang how about [CWE-75](https://cwe.mitre.org/data/definitions/75.html)?

elarlang:
[CWE-624 Executable Regular Expression Error](https://cwe.mitre.org/data/definitions/624.html)

The product uses a regular expression that either (1) contains an executable component with user-controlled inputs, **or (2) allows a user to enable execution by inserting pattern modifiers**.

elarlang (approved):

PR #1670, opened 2023-06-15
Add a requirement about symlinks in zip files includling an example reference to resolve #1424

### LINKED ISSUE #1424: Remove symlinks in zip files
I was checking [this](https://hackerone.com/reports/1439593) HackerOne report with a $29000 bounty and I found it very interesting. This is different than Zip Slip. In case of Zip Slip we can inject .. in the file path so we can extract our file in a wrong place. In this report, the attacker crafts a malicious symlink to /etc/passwd when the backend extracts it untar_zxf function only changes the permissions and extract the symlink as is, so the attacker was able to read the passwd file!
I believe we need an ASVS item to check for removal of symlinks in the zip files.

tghosth:
This Pull Request relates to issue #1424


elarlang (commented):
Do we actually want to link to pen-test reports or blog-style sources?

tghosth:
We have done this in [certain](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x18-V10-Malicious.md#references) [places](https://github.com/OWASP/ASVS/blob/0f9deac4d38bf352de8a891add8acd339265c300/5.0/en/0x13-V5-Validation-Sanitization-Encoding.md?plain=1#L116) already.

elarlang (approved):
I don't think linking to external blogs is good for the standard, but let's get the information in at the moment.

Just for note, alternative CWE candidate https://cwe.mitre.org/data/definitions/61.html

Issue #1610, opened 2023-04-08
Refresh ISO 27001 Multisession Control Statement.


cmlh:
["1.6 Compliance" of MVSP](https://github.com/vendorsec/mvsp/blame/master/packages/mvsp/mvsp.en.asciidoc#LL104) mandates `* Comply with all industry security standards relevant to your business such as PCI DSS, HITRUST, ISO27001, and SSAE 18`.

The parent of all MVSP issues is #1151.

[V3.7 Defenses Against Session Management Exploits](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x12-V3-Session-management.md#v37-defenses-against-session-management-exploits) states "_Previously, based on ISO 27002 requirements, the ASVS has required blocking multiple simultaneous sessions. Blocking simultaneous sessions is no longer appropriate, ..._".

ISO 27002 was updated during 2022 and therefore this statement in ASVS should reflect the latest release of ISO 27002.

I don't know if this is reflect in the latest ISO 27002 or not as I don't have it at hand at the moment.

This issue should also reconsidered when undertaking QA of each future release of ASVS.

tghosth:
In 5.0, I am expecting we will need to trim down this text as much as possible anyway.

tghosth:
Closing as I don't think we will take action on this

tghosth:
Highly likely that this sentance will be removed in 5.0 anyway

tghosth:
I am reopening this and marking it to be considered when we create the actual 5.0 draft.

ryarmst:
@cmlh Related discussion: #2101

PR #1672, opened 2023-06-26
Remove Unknown ISO 27001 Multisession Control

cmlh:
Remove Unknown ISO 27001 Multisession Control.

This Pull Request relates to issue #1610.


tghosth:
I have reopened #1610 but I don't want to do anything about it until we have the 5.0 draft.

Issue #1611, opened 2023-04-11
Insert NIST "Guidelines for Media Sanitization" within IoT Appendix


cmlh:
["1.8 Data handling" of MVSP](https://github.com/vendorsec/mvsp/blame/master/packages/mvsp/mvsp.en.asciidoc#L50) mandates `Ensure media sanitization processes based on NIST SP 800-88 or equivalent are implemented`

The parent of all MVSP issues is #1151.

[Appendix C: Internet of Things Verification Requirement C25](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x93-Appendix-C_IoT.md#security-verification-requirements) states "Verify that the device wipes firmware and sensitive data upon detection of tampering or receipt of invalid message._".

Therefore, modify [Appendix C: Internet of Things Verification Requirement C25](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x93-Appendix-C_IoT.md#security-verification-requirements) to cite [NIST Special Publication (SP) 800-88, Revision 1, Guidelines for Media Sanitization](https://www.nist.gov/privacy-framework/nist-sp-800-88)

mgargiullo:
I wish the ISVS hadn't stalled

tghosth:
Closing as I don't think we will be modifying IoT section

tghosth:
The IoT section was more as a convenience to anyone who wants to go and create a more detailed standard and I am not convinced we will include it in 5.0 anyway

tghosth:
I have opened #1688 to track whether we deprecate this or not.

PR #1673, opened 2023-06-26
Deprecate IoT Appendix for ISVS

cmlh:
Deprecate IoT Appendix for ISVS.

https://github.com/OWASP/ASVS/issues/1611#issuecomment-1597379934
> The IoT section was more as a convenience to anyone who wants to go and create a more detailed standard and I am not convinced we will include it in 5.0 anyway

This Pull Request relates to issue #1611


tghosth:
I want to leave this for now. When we do the rewrite, we will decide finally about this and we need to remember to include an explanation in the final narrative. I have opened https://github.com/OWASP/ASVS/issues/1688 to track whether we deprecate this or not.

Issue #1525, opened 2023-01-24
13.3.2 is too specific, kinda obsolete and effectively duplicates 13.2.6, suggest removal.


tghosth:
Requirement history:

| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **13.3.2** | Verify that the message payload is signed using WS-Security to ensure reliable transport between client and service. | | ✓ | ✓ | 345 |
| **[13.3.2](https://github.com/OWASP/ASVS/commit/b29ed7628b5fed80c8a9c593fe1fa311d1d42a35)** | Verify that the message payload is signed to ensure reliable transport between client and service, using JSON Web Signing or WS-Security for SOAP requests. |  | ✓ | ✓ | 345 |
| **[18.7](https://github.com/OWASP/ASVS/blame/7d1a0c7dd377734cbbd223d9f0e9019b9644084e/4.0/en/0x23-V18-API.md)** | Verify that the message payload is signed to ensure reliable transport between client and service, using JSON Web Signing or WS-Security for SOAP requests. |  | ✓ | ✓ | 345 |

18.7 appeared in ASVS 3.0.1 as 18.9.

Requirement [13.2.6](https://appsecg.host/13.2) (also CWE 345) currently says:

> Verify that the message headers and payload are trustworthy and not modified in transit. Requiring strong encryption for transport (TLS only) may be sufficient in many cases as it provides both confidentiality and integrity protection. Per-message digital signatures can provide additional assurance on top of the transport protections for high-security applications but bring with them additional complexity and risks to weigh against the benefits.

There is [currently a proposal](https://github.com/OWASP/ASVS/issues/1207#issuecomment-1374874043) in #1207 to change it to:

> [MODIFIED] Verify that per-message digital signatures are used to provide additional assurance on top of transport protections for requests/transactions which are highly sensitive or which traverse a number of systems without an uninterrupted TLS connection.

As such, I would propose just dropping this requirement entirely.

elarlang:
Why it obsolete? (I don't know the topic)

By description I agree it is duplicate of 13.2.6

jmanico:
I like this change!

tghosth:
I have opened #1685 to resolve this. This also moves it to section 13.1 because I think this is more generic then just a REST requirement.

Issue #1207, opened 2022-02-02
13.2.6 needs wordsmithing


jmanico:
13.2.6 | Verify that the message headers and payload are trustworthy and not modified in transit. Requiring strong encryption for transport (TLS only) may be sufficient in many cases as it provides both confidentiality and integrity protection. Per-message digital signatures can provide additional assurance on top of the transport protections for high-security applications but bring with them additional complexity and risks to weigh against the benefits.
-- | --

Sjord:
See also [V13.2.6: signing requests · Issue #528 · OWASP/ASVS](https://github.com/OWASP/ASVS/issues/528)

tghosth:
@set-reminder 3 weeks look at this

tghosth:
Ok so I'd like to bring @jaxley back in for his thoughts but I think we need to make this a level 3 requirement as doing anything on top of TLS is rare and complicated.

The requirement as it is seems to basically use TLS as a get out so is probably not super useful anyway.

I also think that even for L3 absolutely requiring message signing is a little extreme so I would be tempted to go with the following wording which at least provides a get out if they have specifically considered it.

| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---:| :---: | :---: |
| **13.2.6** | [MODIFIED] Verify that per-message digital signatures are used to provide additional assurance on top of transport protections or that this was formally considered and rejected (with the rationale documented) based on the additional complexity and risks compared to the benefits. | | | ✓ | 345 |

What do people think?

elarlang:
Against what attack it is and what problem it solves (which is not already solved by using TLS)?

Sjord:
> or that this was formally considered and rejected

I think developers that use ASVS can formally reject any ASVS requirement, if they think it does not apply to them or covers an acceptable risk. This requirement doesn't seem special in that way.

> based on the additional complexity and risks compared to the benefits.

If we expect people to evaluate this, we need to be very clear about the risks and benefits of this requirement. Currently the benefit is stated as "provide additional assurance", which is a bit vague.

> Verify that per-message digital signatures are used

Would client-side TLS count as per-message digital signatures? Would per-message digital signatures offer any benefit on top of TLS?

One situation where per-message digital signatures can be useful, is if the message traverses several systems or is stored in a database. A digital signature can then be verified by the final processor of the message. This provides security over the whole chain, where TLS only provides security between nodes.

tghosth:
[...]

PR #1685, opened 2023-07-10
Fix-13.x.x

tghosth:
This Pull Request relates to issues #1525 and #1207


elarlang (changes requested):
For 13.1.6 please use the label: `MODIFIED, MOVED FROM 13.2.6, LEVEL L2 > L3`

tghosth:
ok is that better @elarlang ?

elarlang:
> ok is that better @elarlang ?

yes, but keep `[ .. ]`

tghosth:
Ugh sorry should be better now 🤦‍♂️🤦‍♂️🤦‍♂️

elarlang (approved):

PR #1689, opened 2023-07-12
Clarify wording of 2.2.4 to resolve #1340

### LINKED ISSUE #1340: Clarify 2.2.9 (and also 2.2.4?)
Based on discussion in #1060, @maizuka made a change to 2.2.4 and 2.2.9 (in #1062).

These now look like this:

| **2.2.4** | [MODIFIED, SPLIT TO 2.2.9] Verify that a hardware-based authenticator and an authenticator that provides verifier impersonation resistance against phishing attacks are used, such as a multi-factor cryptographic hardware, or a combination of a single-factor cryptographic hardware and an OTP device. | | | ✓ | 308 | 5.2.5 |
| **2.2.9** | [ADDED, SPLIT FROM 2.2.4] Verify that multi-factor authentication is required, that is, the application uses either a multi-factor authenticator or a combination of single-factor authenticators. | | ✓ | ✓ | 308 | 4.2.1 |

I like the simplicity of 2.2.9 but I think 2.2.4 is still a little bit clunky. The reason for that seems to be that the NIST guidance is not exactly straightforward but I think the examples we provide are a little confusing.

Also, @RafaelGreen1 suggested in #1287 that we should explicitly mention WebAuthn. 

I think we should reword and simplify 2.2.4 to be:

| **2.2.4** | [MODIFIED, SPLIT TO 2.2.9] Verify that a hardware-based authenticator and an authenticator that provides verifier impersonation resistance against phishing attacks **(such as WebAuthn)** are used. | | | ✓ | 308 | **4.3.1** |

@jmanico what do you think?

tghosth:
This Pull Request relates to issue #1340


elarlang (changes requested):
we don't use markdown syntax in requirement text for styling (`**`)

tghosth:
Yep you are 100% correct, thanks! How about now?

jmanico (approved):

elarlang (approved):

Issue #1674, opened 2023-07-03
8.2.1 - mention or require cache-control no-store


elarlang:
| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **8.2.1** | Verify the application sets sufficient anti-caching headers so that sensitive data is not cached in modern browsers. | ✓ | ✓ | ✓ | 525 |

Is there any actual alternative for `Cache-Control: no-store`?

If there is not, I recommend to at least mention this as an example.

The problem to solve: I see often `Cache-Control: no-cache`, but interpretation of the value `no-cache` can be misleading here.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#no-cache

> Note that no-cache does not mean "don't cache". no-cache allows caches to store a response but requires them to revalidate it before reuse. If the sense of "don't cache" that you want is actually "don't store", then no-store is the directive to use.

`no-store` should be used when response contains sensitive data.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#no-store

> The no-store response directive indicates that any caches of any kind (private or shared) should not store this response.

elarlang:
... and like usual - 0 arguments. What is the reason to have requirement for `no-cache` in ASVS?

elarlang:
... and can you explain me, how `no-cache` value in `Cache-Control` HTTP response header prevents CSRF attack?

elarlang:
Your comments are technically incompetent.

Do you actually understand that `no-cache` does not disallow caching which is quoted in the issue and which is the main point of the issue?

elarlang:
No. My comment is valid.

Additionally you should understand the difference between cache for middleware (which you are referring) and the cache configuration for browser (which is the issue here).

elarlang:
> no-cache prevents CSRF by forcing the web browser to revalidate the Synchronizer Token with the web server before sending the POST request.

And you don't dare here to explain, how and why this `no-cache` actually works in this process? I _assume_ (which is in general bad thing to do) that there is extra HTTP request to achieve fresh tokens before data-changing request (in your example HTTP request with POST method). But it's not possible to read it out from your line - how browser is revalidating it.

[...]

PR #1699, opened 2023-07-20
#1674, no-store

elarlang:
This Pull Request relates to issue #1674


elarlang:
We have separate requirement for cache (8.2.1) and separate requirement  for PII in general (8.2.2).

You prove again and again my previous comment (https://github.com/OWASP/ASVS/issues/1674#issuecomment-1623132011) - the context here is HTTP response header and you copy-paste quotes from Request headers.

tghosth (approved):

Issue #1299, opened 2022-06-21
new requirement - one application per hostname (and remove current 3.4.5)


elarlang:
Initial discussion on requirement 3.4.5 in issue #978 . But I would solve it a bit different way and it deserves separate issue, as it is quite big change.

Problems to solve: if one application is compromised, it should not affect others. Even those are separated on the filesystem or hosting, host, origin or domain name will be the keyword here.

If the application uses cookies
* we need to have one application per host/origin and `__Host-` prefix used for session cookies
* OR we need to have one application per eTLD (effective Top Level Domain)

If the appication do not use cookies
* we need to have separation for SOP (Same Origin Policy)

If this idea develops as new requirement, it drops current 3.4.5.

Sjord:
> we need to have one application per host/origin

This should be "host", not origin. Specifically, it should not be allowed to run different applications on different ports. Applications running on different TCP ports would have different origins, but the `__Host-` prefix does not protect cookies on different ports from interfering.

> we need to have one application per eTLD (effective Top Level Domain)

While this makes sense in respect to cookies, it makes phishing much easier. Checking whether the domain name is legit is easier for online.microsoft.com than for microsoftonline.com.

I would propose to have a requirement saying "every application must run on its own (sub)domain", and not make distinction between cookies or not, and not mention cookie prefixes as it's already a requirement.

elarlang:
> Applications running on different TCP ports would have different origins

technically possible, never seen in practice for public websites. It is used for internal API's, probably we need to exclude this somehow in the requirement?

> we need to have one application per eTLD (effective Top Level Domain)

this was given as an alternative. If there is just one application per domain, we don't have a lot of other problems and therefore the need to solve them. But I agree, it's easier to leave this option completely out

> every application must run on its own (sub)domain

I think host**name** is correct to user here instead of domain or subdomain

tghosth:
@elarlang 
Does this requirement boil down to something like:

> Verify that multiple, separate applications are not hosted on the same hostname.

elarlang:
For me, yes. I'm a bit concerned, is it clear enough. Like how to define, what is an application? If there is "public site", "self service" and "back office" - are those part of the same application or are those separate applications.

tghosth:
Maybe the answer is applications with separate session mechanisms should be on different origins?

[...]

PR #1701, opened 2023-07-27
Address #1299 by changing 3.4.5 into an architecture level requirement.

tghosth:
This Pull Request relates to issue #1299


elarlang:
See https://github.com/OWASP/ASVS/issues/1299#issuecomment-1654005638

jmanico (approved):

elarlang:
We are going to add the requirement, but not to the category like proposed in the PR. We open separate PR when it finds it's logical place. Discussion in the issue #1299

PR #1710, opened 2023-09-07
Return 2.1.12 with a more sensible modification to resolve #1265

### LINKED ISSUE #1265: Requirement of masking password
The removing of [2.1.12](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x11-V2-Authentication.md#v21-password-security) was taken at #570. At the same time, it has also removed a requirement for typed passwords to be masked.

> | **2.1.12** | Verify that the user can choose to either temporarily view the entire masked password, or temporarily view the last typed character of the password on platforms that do not have this as built-in functionality. | ✓ | ✓ | ✓ | 521 | 5.1.1.2 |

Should we cover this as a new requirement, or is it no longer needed?

My suggestion:
> Verify that passwords are not displayed as plain text in pages and password fields. Input fields should use type=password.

tghosth:
This Pull Request relates to issue #1265


elarlang:
see https://github.com/OWASP/ASVS/issues/1265#issuecomment-1709915556

elarlang (changes requested):
I think CWE-549 suits better: CWE-549 Missing Password Field Masking

tghosth:
Nice! I did not know there was a CWE like that!

elarlang (approved):

PR #1736, opened 2023-09-28
Adding 5.2.11 to resolve #1533

### LINKED ISSUE #1533: JNDI injection prevention checks should be added to the ASVS list
The application should prevent JNDI injection attacks by properly validating and sanitizing user input before using it in a JNDI query. (We have items for SQL injection, LDAP injection, SMTP injection,... but we don't have an item for JNDI injection)

Rationale:

JNDI injection is a type of attack that occurs when an attacker is able to inject malicious data into the JNDI data source used by an application, allowing them to execute arbitrary code on the application server. Proper validation and sanitization of user input can prevent this type of attack by ensuring that only safe data is used in JNDI queries.

Remediation:

Validate user input by checking that it conforms to the expected format, length, and range of values.
Use parameterized JNDI queries to prevent malicious data from being interpreted as part of the query.
Sanitize user input by removing or encoding special characters that could be used to inject malicious data into the JNDI query.
Review and test the application's input validation and sanitization mechanisms to ensure they are properly implemented and effective.

Testing Method:

- Verify that the application properly validates and sanitizes the input before using it in the JNDI query.

References:

[Exploiting JNDI Injections](https://www.veracode.com/blog/research/exploiting-jndi-injections-java)

tghosth:
This Pull Request relates to issue #1533


elarlang (changes requested):
`| **5.2.11** | | [ADDED]` - one pipe too much

elarlang (approved):
fixed it myself

PR #1764, opened 2023-10-22
SKF link is no longer valid and is causing an error on the link verifier

tghosth:
@elarlang please merge :)


elarlang (approved):
Just merge, I did not validate the content - are the content there actually aligned with the ASVS

Issue #1481, opened 2022-12-27
CWE mapping will be dropped


elarlang:
I think CWE mapping is not useful/valuable at the moment. Sometimes it's useful to validate, is it correlating with ASVS requirement text but in big picture - is this mapping actually used?

At the moment we have mapped requirements to obsoleted categories, to views, etc.

[CWE-16](https://cwe.mitre.org/data/definitions/16.html) section Notes:
> Use for Mapping: Prohibited (this CWE ID must not be used to map to real-world vulnerabilities). Categories are informal organizational groupings of weaknesses that help navigation and browsing by CWE users, but they are not weaknesses in themselves [REF-1287]. ... Mapping is also Prohibited because this entry's status is Obsolete.

List of ASVS requirements where mapping points to CWE which type is `view` or `category` (not `weakness`) and/or status is obsolete

* [CWE-16](https://cwe.mitre.org/data/definitions/16.html) status:obsolete; type:category; ASVS: 2.5.4, 3.4.4, 3.4.5, 10.3.1, 14.1.3
* [CWE-19](https://cwe.mitre.org/data/definitions/19.html) status:draft; type:category; ASVS: 8.1.5, 8.1.6
* [CWE-255](https://cwe.mitre.org/data/definitions/255.html) status:draft; type:category; ASVS: 2.10.2
* [CWE-265](https://cwe.mitre.org/data/definitions/265.html) status:incomplete; type:category; ASVS: 1.14.5, 14.2.6
* [CWE-275](https://cwe.mitre.org/data/definitions/275.html) status:draft; type:category; ASVS: 1.4.5
* [CWE-310](https://cwe.mitre.org/data/definitions/310.html) status:draft; type:category; ASVS: 2.6.3, 2.7.6, 6.2.1
* [CWE-320](https://cwe.mitre.org/data/definitions/320.html) status:obsolete; type:category; ASVS: 1.6.1, 1.6.2, 1.6.3, 1.6.4, 2.8.2, 2.9.1, 6.4.2
* [CWE-1002](https://cwe.mitre.org/data/definitions/1002.html) status:incomplete; type:category; ASVS: 14.1.6
* [CWE-1008](https://cwe.mitre.org/data/definitions/1008.html) status:incomplete; type:view; ASVS: 9.4.4
* [CWE-1009](https://cwe.mitre.org/data/definitions/1009.html) status:draft; type:category; ASVS: 1.7.1
[...]

Issue #1470, opened 2022-12-22
proposal: remove current 5.3.9 as it is in incorrect place and covered by other requirements


elarlang:
Current requirement 5.3.9

| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **5.3.9** | Verify that the application protects against Local File Inclusion (LFI) or Remote File Inclusion (RFI) attacks. | ✓ | ✓ | ✓ | 829 |

[CWE-829](https://cwe.mitre.org/data/definitions/829.html): Inclusion of Functionality from Untrusted Control Sphere

I try to understand, why requirement 5.3.9 is in category "V5.3 Output Encoding and Injection Prevention" and also, why it exists at all. We may watch it as "path or url injection attack", but defence method here is not encoding or escaping, it must be validation and sanitization.

It's also not clear, is it addressing only PHP style file inclusion which also means that included file is executed as PHP program code. Or is it covering also just file read?

Chosen CWE points more towards V10 category https://github.com/OWASP/ASVS/issues/1471

In "V12.3 File Execution" we have requirements:

| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **12.3.1** | Verify that user-submitted filename metadata is not used directly by system or framework filesystems and that a URL API is used to protect against path traversal. | ✓ | ✓ | ✓ | 22 |
| **12.3.2** | Verify that user-submitted filename metadata is validated or ignored to prevent the disclosure, creation, updating or removal of local files (LFI). | ✓ | ✓ | ✓ | 73 |
| **12.3.3** | Verify that user-submitted filename metadata is validated or ignored to prevent the disclosure or execution of remote files via Remote File Inclusion (RFI) or Server-side Request Forgery (SSRF) attacks. | ✓ | ✓ | ✓ | 98 |

* [CWE-22](https://cwe.mitre.org/data/definitions/22.html): Improper Limitation of a Pathname to a Restricted Directory ('Path Traversal')
* [CWE-73](https://cwe.mitre.org/data/definitions/73.html): External Control of File Name or Path
* [CWE-98](https://cwe.mitre.org/data/definitions/98.html): Improper Control of Filename for Include/Require Statement in PHP Program ('PHP Remote File Inclusion')

[...]

Issue #1427, opened 2022-11-17
user-submitted filename metadata (V12.3)


Sjord:
[ASVS/0x20-V12-Files-Resources.md at master · OWASP/ASVS · GitHub](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x20-V12-Files-Resources.md)

| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **12.3.1** | Verify that user-submitted filename metadata is not used directly by system or framework filesystems and that a URL API is used to protect against path traversal. | ✓ | ✓ | ✓ | 22 |
| **12.3.2** | Verify that user-submitted filename metadata is validated or ignored to prevent the disclosure, creation, updating or removal of local files (LFI). | ✓ | ✓ | ✓ | 73 |
| **12.3.3** | Verify that user-submitted filename metadata is validated or ignored to prevent the disclosure or execution of remote files via Remote File Inclusion (RFI) or Server-side Request Forgery (SSRF) attacks. | ✓ | ✓ | ✓ | 98 |
| **12.3.5** | Verify that untrusted file metadata is not used directly with system API or libraries, to protect against OS command injection. | ✓ | ✓ | ✓ | 78 |

I think "filename metadata" is incorrect here. It would be information about the filename, not the filename itself.

I think "filename metadata" should be either "filename" or "file metadata". I would prefer "file metadata", as that would include both the filename and content-type, and any other file information sent by the client.

elarlang:
Oh, I have postponed to reorganize this section as I have some big tasks here opened already, but in this section are so many things which are incorrect or I just don't like (https://github.com/OWASP/ASVS/issues/1136#issuecomment-974680885). For a starter:
* user-submitted - unnecessary limitation, should be for every filename or metainfo an application processes
* in general we should just say - filename and metainfo should be considered as user input and we can use other requirement in V5 category

elarlang:
I think this entire "#V12.3 File Execution" should be moved away from V12 - all other requirements in V12 are more related to handling user-uploaded or application generated files as content.

V12.6.1 is already waiting to be moved away but there is no clear-good category yet (#1343).

From V12.3:
* 12.3.1 path traversal - should be in validation + sanitization
* 12.3.2 local file - the point is the same with path traversal, it should be not possible to manipulate the path where the resource is created/read/deleted
[...]

PR #1776, opened 2023-11-01
chore: Update 12.3.3 CWE mapping from CWE-98 to CWE-73

hazanasec:
Update the mapping of the ASVS requirement 12.3.3. The current mapping is to [CWE-98](https://cwe.mitre.org/data/definitions/98.html), which is specific to PHP's "Include/Require" statements. The proposed change is to map it to [CWE-73](https://cwe.mitre.org/data/definitions/73.html), which is a more generalized weakness that accurately describes the problem of allowing user-submitted input to control or influence file paths or names in filesystem operations. This more inclusive mapping covers a broader range of technologies and scenarios, making it a more suitable choice.


tghosth:
Seems close enough to me, @elarlang any objections?

elarlang:
I agree to make the change as it does not change anything meaninful (as we most likely going to drop CWE mapping - https://github.com/OWASP/ASVS/issues/1481), but I don't like the process here. I also don't think this requirement survives the same way to v5.0.

We have related issues opened for the topic (https://github.com/OWASP/ASVS/issues/1470, https://github.com/OWASP/ASVS/issues/1427).

Usual process: we open issues for having discussion and agreement for the proposed change and if we have agreement, then we go for PR.

Ok, here the change is so small that issue-process may feel like overhead. On the other hand, this is the way we would like to keep overview and reasoning for the future.

tghosth (approved):

tghosth:
I think this is small enough and self-explanatory enough to make an exception but I agree we would usually prefer an issue first.

PR #1779, opened 2023-11-08
Update 0x22-V14-Config.md

andreas-becker:
This Pull Request relates to no issue, because it only fixes a small typo.


elarlang:
@andreas-becker - if you want to propose some changes, please follow "issue first" > "agreement in issue" > "PR flow like described here: https://github.com/OWASP/ASVS/blob/master/CONTRIBUTING.md#before-you-open-a-pull-request

This current proposal is not correct, as header name is `Referer` not `Referrer`. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer

So I'm closing the PR.

PR #1784, opened 2023-11-09
Include privacy-enhancing technologies in 1.8.2 to resolve #1752

### LINKED ISSUE #1752: Consider Adding Checks for Privacy Preserving Schemas in ASVS
While reviewing the ASVS, I've observed a potential gap: there isn’t a specific check or item considering the use of privacy-preserving schemas. In an era where data privacy and protection is paramount, it's crucial for sensitive applications to incorporate methods like differential privacy, Zero-Knowledge Proofs (ZKP), and other privacy-preserving techniques.

To illustrate the significance of such techniques, let's consider the use case of homomorphic encryption, specifically the Paillier cryptosystem. Imagine an application needs to compute the average of a series of data points, but the entity doing the computation should not have access to the actual raw data due to privacy concerns. Using the Paillier cryptosystem, this can be achieved as the system allows for specific computations on ciphertexts which, when decrypted, match the result of the operations as if they were performed on the plaintext.

Without getting too deep into the intricacies, here's a basic overview:

- Data providers encrypt their data points using a public key.
- The application, without ever decrypting the data, computes the average on the encrypted values.
- The result is decrypted by an entity with a private key to obtain the actual average, ensuring data privacy throughout the process.

By incorporating such a check in the ASVS for privacy-preserving techniques, we'd be offering guidelines for developers and security professionals to ensure that sensitive data operations maintain user privacy.

tghosth:
This Pull Request relates to issue #1752

### DIFF
--- a/5.0/en/0x10-V1-Architecture.md
+++ b/5.0/en/0x10-V1-Architecture.md
@@ -89,7 +89,7 @@ Architectural requirements are intrinsic to the entire code base, and thus diffi
 | # | Description | L1 | L2 | L3 | CWE |
 | :---: | :--- | :---: | :---: | :---: | :---: |
 | **1.8.1** | [MODIFIED, MERGED FROM 8.3.4, LEVEL L2 > L1] Verify that all sensitive data created and processed by the application has been identified and classified into protection levels, and ensure that a policy is in place on how to deal with sensitive data. | ✓ | ✓ | ✓ | 213 |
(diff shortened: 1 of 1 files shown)


elarlang (approved):
ntoe for the future - if we limit V1 to documentation requirements, then we need to have separate requirement to cover "and that these are applied in the architecture"

tghosth:
> ntoe for the future - if we limit V1 to documentation requirements, then we need to have separate requirement to cover "and that these are applied in the architecture"

I imagine this will be covered in the V1 rework.

PR #1840, opened 2024-01-24
Clarify the language in 2.10.2 to resolve #1521

### LINKED ISSUE #1521: 2.10.2 needs rewording
| **2.10.2** | [GRAMMAR] Verify that if passwords are required for service authentication, the service account used is not a default credential (e.g. root/root or admin/admin are default in some services during installation). | | ✓ | ✓ | 255 | |

As I understand it, this requirement is related to a case where 2.10.1 cannot be followed for some reason. We should make this clearer in some way but only once #1032 which changes 2.10.1 is closed.

@set-reminder 8 weeks @tghosth has #1032 been closed yet?

tghosth:
This Pull Request relates to issue #1521


tghosth (commented):
I think the GRAMMAR tag is sufficient rather than MODIFIED because the meaning of the requirement does not change.

elarlang (approved):

Issue #810, opened 2020-06-16
Cleanup verification descriptions from links


elarlang:
## Links

There are also 77 requirements with links to <https://owasp.org/www-project-proactive-controls/#div-numbering>

```
$ grep -r "\[C" . | grep "\*\*" | wc -l
77
```

Those links are also not part of requirement text, maybe we should have another column in tables for them?

Problem (like described in #809) - if to use data in JSON format or want to use description text, then you need to cleanup requirement texts manually. Description text is not usable like it should be.

tghosth:
I deliberately want these particular links in there. Taking discussion to #777 for now.

elarlang:
74 of 320 requirements have this link. Maybe we should have this link in separate column in table like CWE and NIST, then it is not part of description text.

NIST reference is set to 69 requirements.

tghosth:
I don't want to create too many columns...

tghosth:
Reassigning this issue to @vanderaj as it kind of feeds into how we generate the different formats (makefile and #777) and if the existence of lots of links in the document makes a mess for things like the JSON or CSV formats

jmanico:
Having a link inside of a requirement is a very bad idea. That means the requirement is not good enough as is, which is a problem. All links inside of requirements need to go in the summary at the bottom of the section if they are needed.

This is a really big deal that we fix this, in my opinion.

jmanico:
> I don't want to create too many columns...

I agree @tghosth but links inside of requirements are a very bad idea. Can we move those links to "references" under each section with your blessing?

tghosth:
These are not links to references, these are links to the relevant proactive control entry. It is more like a mapping. I think we need to agree a standard mapping format and maybe then move these to a different format.

jmanico:
I strongly disagree Josh, I would like to take any links or references out of the requirement text itself, it means the requirement is not good enough as is. We can move the links to other parts of ASVS, just not in the requirement text itself.

tghosth:
I think we do in fact agree :) If these were references for more information like in #664 then we should remove them (as I did in #664.)

In this case, these are not references for more information but rather just a mapping to track which requirements are in which proactive control. I also think that the mapping should be removed but we don't have a better solution for that yet, I have left the issue open so that we can deal with this at some point :)

jmanico:
Ok cool. Can we add a new column to the ASVS for them? Like instead of

[...]

Issue #809, opened 2020-06-16
Cleanup verification descriptions (V3.4.4 and V13.2.3) from references


elarlang:
## References

[V3.4.4](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x12-V3-Session-management.md#v34-cookie-based-session-management)

> V3.4.4 | Verify that cookie-based session tokens use "__Host-" prefix **(see references)** to provide session cookie confidentiality.

[V13.2.3](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x21-V13-API.md#v132-restful-web-service-verification-requirements)

> V13.2.3 | Verify that RESTful web services that utilize cookies are protected from Cross-Site Request Forgery via the use of at least one or more of the following: triple or double submit cookie pattern **(see [references](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html))**, CSRF nonces, or Origin request header checks.

Only 2 requirements have "see references" as part of requirement text and only one of them is provided in markdown syntax link.

I think requirement description should have clean text without any extra references.

Problem - if to use data in JSON format or want to use description text, then you need to cleanup requirement texts manually. Description text is not usable like it should be.

tghosth:
I agree with avoiding links within the requirement text. I will take this as there is another issue related to triple submit cookies.

elarlang:
This related question can be based on my created issue there: https://github.com/OWASP/CheatSheetSeries/issues/195

Another question - V13.2.3 is a bit duplicate for V4.2.2 where there is no list of potential defense solutions. Isn't this list actually too complicated to maintain - to list all potential solutions?

tghosth:
What is in 4.2.2?

elarlang:
[V4.2.2](https://github.com/OWASP/ASVS/blob/master/4.0/en/0x12-V4-Access-Control.md#v42-operation-level-access-control)

> V4.2.2 Verify that the application or framework enforces a strong anti-CSRF mechanism to protect authenticated functionality, and effective anti-automation or anti-CSRF protects unauthenticated functionality.

Another related issue to V4.2.2 - #795

tghosth:
OK, taking your other question to #795.

Action on me to get the triple submit cookie thing resolved.

tghosth:
Backported to v4.0.2 by #822

PR #1877, opened 2024-02-26
Remove proactive controls from requirement text

Sjord:
As agreed in #810

See also #847, #809


tghosth:
@Sjord any chance you could create a separate file of the mapping (v4.0.3 requirement ID to Proactive Controls #) so that we can continue to have it somewhere if necessary?

elarlang:
> @Sjord any chance you could create a separate file of the mapping (v4.0.3 requirement ID to Proactive Controls #) so that we can continue to have it somewhere if necessary?

We have carried the values from v4.0.3 and we have not added or mapped any new requirements or changed requirements. I think the mapping is relevant only for v4.0.3.

I prefer just to get rid of them or those should be covered via the mapping project.

tghosth (approved):

PR #1931, opened 2024-04-18
Resolve #1918 by clarifying 8.3.x requirements and making L3

### LINKED ISSUE #1918: V8.3 security and policy
| **8.3.2** | [MODIFIED, SPLIT TO 8.3.9] Verify that users have a method to remove their data on demand. | ✓ | ✓ | ✓ | 212 |
| **8.3.3** | Verify that users are provided clear language regarding collection and use of supplied personal information and that users have provided opt-in consent for the use of that data before it is used in any way. | ✓ | ✓ | ✓ | |
| **8.3.9** | [ADDED, SPLIT FROM 8.3.2] Verify that users have a method to export their data on demand. | | ✓ | ✓ | |

Can we say for each requirement, what is the risk to solve? Is it worth a requirement in the ASVS or those are here because they are just from GDPR or similar?

If we want to keep them:

  * should we require level 1? 8.3.2 level 1 vs 8.3.9 level 2 do not match well.
  * maybe we should have separate policy related/oriented section
  
  edit: cwe 212 is kind of related, but not correct

tghosth:
This Pull Request relates to issue #1918

### DIFF
--- a/5.0/en/0x12-V3-Session-management.md
+++ b/5.0/en/0x12-V3-Session-management.md
@@ -90,7 +90,7 @@ Attackers start a half-open attack by attempting to lock, reset, or recover a cr
 
 | # | Description | L1 | L2 | L3 | CWE | [NIST &sect;](https://pages.nist.gov/800-63-3/sp800-63b.html) |
 | :---: | :--- | :---: | :---: | :---: | :---: | :---: |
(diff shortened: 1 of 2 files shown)


elarlang (changes requested):
For 8.3.3
* "uses of personal" vs " uses personal"?
* as it got modified and level changed, tag `[MODIFIED, LEVEL L1 > L3]` (edit: L1)

elarlang (approved):

Issue #1938, opened 2024-04-23
Add ReqView format as generated output


rfricz:
We’ve written the code to generate the ASVS as a document template for [ReqView](https://www.reqview.com/) requirements management tool. The added value is that you can customize and trace security requirements throughout the product development lifecycle.

The intended usage of this document template is to *define* and *implement* security requirements comprehensively as described at [OWASP Top Ten Proactive Controls 2018 – C1: Define Security Requirements](https://owasp.org/www-project-proactive-controls/v3/en/c1-security-requirements). The user would link the ASVS security requirements to User Stories (Needs), Use/Misuse Cases, SW Requirements Specification (SRS), Risks, Tests etc.

We’ve incorporated the ASVS document in the [*Example project*](https://www.reqview.com/doc/example-project/) included in [ReqView webapp](https://desktop.reqview.com/). Detailed information about the template can be found at [ReqView documentation](https://www.reqview.com/doc/asvs-template/).

ReqView makes managing ASVS requirements really convenient:
* You can find all the explanatory text (Control Objectives, Glossary, References etc.) that applies to the currently focused section, chapter or requirement in the [Instructions pane](https://www.reqview.com/doc/asvs-template/#template-instructions) (resolves #821 in a way).
* The CWE and NIST § columns show clickable links to the corresponding items – no manual searching like when using the PDF version.
* Apart from L1/2/3 columns that are displayed just like in the md docs, you can use additional [attributes](https://www.reqview.com/doc/asvs-template/#attributes) for easier workflow.
* You can switch between user role based *Manage* view, *Traceability* view and dedicated *Compliance* views with predefined per-level filters and additional *Owner*, *Compliance Status* and *Evidence* columns useful to both assessors and people responsible for compliance.
* You can trace ASVS requirements using custom traceability matrix, see the image below.
* You can export fully customizable documentation in HTML, DOCX, XLSX, PDF, ReqIF, or any custom text format.

ReqView [data format](https://www.reqview.com/doc/file-data-format/) is open, JSON-based and documented by JSON schemas. It’s also similar to the JSON that you already generate.

![ASVSTableViewTraceability](https://github.com/OWASP/ASVS/assets/36163339/0b02518a-8118-43ca-99af-8c5f10fa46d8)

elarlang:
Hi.
however useful the tool is, personally I don't think that the ASVS as a free and open-source project should add an output format for the commercial tool. This can be used on the tool side as an adapter.

[...]

PR #1939, opened 2024-04-23
Add ReqView format as generated output

rfricz:
This code adds support for generating ReqView document templates (.reqw) from the source .md files. It’s mostly an extension of the existing JSON generators for both 4.0 and 5.0 (they remain identical), because the structure of the output is very similar to the already generated JSON.

Basically, a template ReqView JSON file is read and filled with data from the .md files. Some additional Markdown processing is done to be able to separate the informative text from the actual requirements’ text, and also HTML is rendered from the processed bits.

The generated outputs for v4.0.3 can be found at [ReqView docs site](https://www.reqview.com/doc/asvs-template/#document-template), current v5.0 is available as [OWASP_Application_Security_Verification_Standard-5_en.zip](https://github.com/OWASP/ASVS/files/15080030/OWASP_Application_Security_Verification_Standard-5_en.zip).

This Pull Request relates to issue #1938, where I’ve tried to explain the use case for generating this format.


tghosth:
Closed as discussed in #1938

Issue #1731, opened 2023-09-26
Are format string vulnerabilities only relevant to memory unsafe languages?


tghosth:
Are format string vulnerabilities only relevant to memory unsafe languages?

Requirement [5.4.2](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x13-V5-Validation-Sanitization-Encoding.md#v54-memory-string-and-unmanaged-code) is in a section that is more about memory unsafe languages but I have two problems:

1. I don't think format string vulnerabilities are only in memory unsafe languages, [for example this article also talks about python](https://www.invicti.com/blog/web-security/format-string-vulnerabilities/).
2. This feels more like an input validation or sanitisation requirement and would seem to fit better there.

Any thoughts?

History:

| # | Description | L1 | L2 | L3 | CWE |
| --- | --- | --- | --- | -- | -- |
| **5.4.2** | Verify that format strings do not take potentially hostile input, and are constant. | | ✓ | ✓ | 134 |
| 5.3.12 | Verify that if the application uses a systems language or unmanaged code, format strings do not take potentially hostile input, and are constant. |  | ✓ | ✓ | 4.0 |

tghosth:
Josh to move this somewhere....

csfreak92:
Hi @tghosth, I was browsing the other issues and our current ASVS 5.0 repo and I think these requirements `5.4.2` and `5.3.12` you have listed above can fit the Defensive Coding part of this section: https://github.com/elarlang/ASVS/blob/master/5.0/en/0x18-V10-Malicious.md#v104-defensive-coding. 

What do you think?

tghosth:
> Hi @tghosth, I was browsing the other issues and our current ASVS 5.0 repo and I think these requirements `5.4.2` and `5.3.12` you have listed above can fit the Defensive Coding part of this section: https://github.com/elarlang/ASVS/blob/master/5.0/en/0x18-V10-Malicious.md#v104-defensive-coding.
> 
> What do you think?

I'm not sure I agree @csfreak92 , this seems like a pretty clear cut validation/sanitization. 10.4 is basically for code level requirements with other types of mitigation that don't fit anywhere else :)

I would suggest putting this in "V5.2 Sanitization and Sandboxing" and rewording as follows:

| # | Description | L1 | L2 | L3 | CWE |
| --- | --- | --- | --- | -- | -- |
[...]

PR #1982, opened 2024-06-21
Update 0x13-V5-Validation-Sanitization-Encoding.md

jmanico:
work for https://github.com/OWASP/ASVS/issues/1731


tghosth:
@csfreak92 is that sufficent extra context do you think?

csfreak92:
It looks good, @tghosth though there's something missing I think. I say that because format string vulnerabilities are kind of like a command injection through these format strings being interpreted as commands. Would sanitization of input be enough? I'm pondering as well if that's enough or not. Welcome for more feedback.

tghosth:
Sanitization should always be enough I think... It generally means removing special characters so should probably be the most aggressive protection

csfreak92:
Ah I see, ok then that closes it. I concur with this PR. :)

jmanico:
Respectfully, sanitation has a history of causing harm and is the last resort. For example if you remove ../ then ….// will bypass. Escaping and/or encoding is always the preferable defense when it comes to injection, expect SQL which is of course querying parameterization. 

I’m sorry to add feedback late in the process.

tghosth (approved):

Issue #1612, opened 2023-04-11
Request for Addition of WebRTC Security Subcategory in ASVS


ImanSharaf:
I noticed that there is a dedicated subcategory for WebSocket security, but not for WebRTC (Web Real-Time Communication).  Given the growing adoption of WebRTC in modern applications for audio, video, and data communication, I believe it is essential to address its security considerations within the ASVS.

I would like to propose the addition of a new subcategory specifically for WebRTC security. This subcategory could include items such as:

- Ensuring secure WebRTC signaling channels (e.g., TLS)
- Validating and sanitizing SDP (Session Description Protocol) exchanges
- Implementing secure WebRTC session termination
- Utilizing secure communication mechanisms for media and data streams such as DTLS-SRTP

I can write them in the ASVS style after we confirm what requirements we want to add.
Also, there are some good write-ups that we can extract valuable checks out of them such as [this one](https://www.rtcsec.com/article/slack-webrtc-turn-compromise-and-bug-bounty/) ($3,500 was paid as bounty by Slack): 
- Verify that the TURN server implements blacklist to block sensitive addresses from being specified as the XOR-PEER-ADDRESS in TURN messages

Inclusion of these items and potentially others would significantly benefit the community by raising awareness and providing guidance on securing WebRTC implementations.

tghosth:
Can you suggest requirements that you think are not explicitly covered by existing requirements.

For example, I think we already have a requirement that all communications should be via TLS so I don't think we need extra requirements for that.

tghosth:
hi @ImanSharaf any thoughts on this?

ImanSharaf:
The specific check related to the TURN server and XOR-PEER-ADDRESS addresses a unique WebRTC-related concern that isn't necessarily covered by general security guidelines. SDP exchanges are unique to WebRTC and similar real-time communication protocols. Validating and sanitizing these exchanges are essential as they contain vital information about the media and data streams. They could be exploited if not secured properly. Unlike traditional communication channels, WebRTC sessions have unique characteristics, and ensuring their secure termination is essential to avoid unintended data leaks or unauthorized access.

ImanSharaf:
Sure, do we want to create a subcategory for WebRTC similar to Websocket?

elarlang:
How many requirements you have in mind? Web Socket category in my opinion contains duplicates which can easily handled with "parent" or more general requirement (but this is separate topic).

tghosth:
[...]

Issue #1875, opened 2024-02-24
6.2.1 causes capitalization inconsistency 


alitasdln:
Other attacks in the document such as "brute force", "rainbow table", "phising" are all in lowercase but in [this requirement](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x14-V6-Cryptography.md#v62-algorithms) "Padding Oracle" is all in uppercase which may cause confusion for people who never heard about it. Especially considering Oracle is a well-established company in technology field. 

My proposal is to write it in all lowercase as other attack methods.

elarlang:
We can change this requirement, but I think we need to set some rules in place and follow proper grammar in the document.

Does anyone have some good grammar rules to point out?

ping @tghosth @jmanico @Sjord

jmanico:
Yes. Use a tool like Grammarly to check your grammar.

elarlang:
> Yes. Use a tool like Grammarly to check your grammar.

not helpful in this case, as it does not know which words are names for something and which are not.

Sjord:
> Padding Oracle

I agree that this should be lower case, and also the following should be lower case:

- libraries do not contain Easter eggs
- This should prevent Web Cache Deception attacks.
- Verify that the application is protected against ... Formula Injection
- to prevent ... Request Smuggling
- Verify that proper certification revocation, such as ... Stapling

I think in these cases uppercase is acceptable:
- JavaScript, LaTeX, Markdown, GraphQL, WebSocket
- CSV, HTTP, SMTP, API
- Content-Type, Set-Cookie

When an abbreviation is written in full it is often capitalized. I think these also should not be capitalized:
- Server-side Request Forgery (SSRF)
- Insecure Direct Object Reference (IDOR)

Currently all requirements start with "Verify that". I would rather omit that.

The first sentence of a requirement is often pretty clear and grammatically correct. The second sentence is often problematic; written for a different audience than the first, or gives a recommendation instead of a requirement:

- Registration and forgot password functionality should also have this protection.
- Strongly consider logging only in UTC if systems are global to assist with post-incident forensic analysis.
- Permissions should still be allocated using roles.

The requirements table should not have markdown formatting, because this doesn't work if it's exported to CSV or another format. Currently the links to proactive controls are the biggest problems here, and I suggest removing them.

Sjord:
[...]

PR #1995, opened 2024-07-24
Create a new chapter for WebRTC.

tghosth:
This Pull Request relates to issue #1612


tghosth:
I would request a final view from @elarlang's discerning eye before we merge this into main :)

elarlang (changes requested):
Acronyms in requirements - avoid or use in style: Web Real-Time Communication (WebRTC)

53.1.1, 53.1.2 - TURN acronym

53.1.2 - is it WebRTC specific? Seems a general requirement and potentially covered by 12.7.1

| **12.7.1** | [ADDED] Verify that the application proactively releases system resources, such as database connections, open files, threads, etc, when it finishes using them to prevent resource exhaustion. | | ✓ | ✓ | 404 |

53.2.1 - seems out of scope, and if not, then it is a general requirement for V9

53.2.2 - "Verify that"? It seems general V9 requirement (not WebRTC specific)

53.2.3 - publically known vulnerable sounds like a duplicate of 14.2.1

| **14.2.1** | Verify that all components are up to date, preferably using a dependency checker during build or compile time. | ✓ | ✓ | ✓ | 1026 |

race condition test is covered by 11.1.6. Is there something WebRTC-specific?

| **11.1.6** | [MODIFIED] Verify that the application uses synchronization and locking mechanisms for sensitive operations in order to keep internal data consistent, maintain user state, and prevent race conditions, such as 'time of check to time of use (TOCTOU)' vulnerabilities. | | ✓ | ✓ | 367 |

53.2.4, 53.2.5 - SRTP, RTP acronyms

53.2.5, 53.2.6 - slash (/) instead of "and" or "or"

53.2.6, 53.2.7, 53.3.1, 53.3.2 - "can handle" is a bit vague. What it means?

I only checked requirement texts (not checked chapter texts).

sandrogauci:
Thanks @elarlang for the checks and feedback! 

> Acronyms in requirements - avoid or use in style: Web Real-Time Communication (WebRTC)

WebRTC is full of acronyms so we cannot really avoid them much.

I have a general question about this: would it be preferred that the acronyms be defined when mentioned first time, or in a section just for acronyms or do we instead define them each time mentioned in a requirement in the indicated format?

Also - is this information about when and how to define acronyms publicly documented?

> 53.1.1, 53.1.2 - TURN acronym
> 
> 53.1.2 - is it WebRTC specific? Seems a general requirement and potentially covered by 12.7.1

This is TURN specific, which is often a critical important of WebRTC infrastructure. While 12.7.1 is about releasing resources and is very generic, our requirement is about being able to handle the situation where the TURN server is asked for relay/proxy a larger number of IP:ports than it can handle. Releasing resources is not what's recommended in this case. TURN servers, in some ways, are similar to web proxies in that they are mean to relay connections or UDP packets.


> #	Description	L1	L2	L3	CWE
> **12.7.1**	[ADDED] Verify that the application proactively releases system resources, such as database connections, open files, threads, etc, when it finishes using them to prevent resource exhaustion.		✓	✓	404
> 53.2.1 - seems out of scope, and if not, then it is a general requirement for V9

I do not think that this is a general requirement because we are specifically writing about DTLS. Please take a look at my replies to @tghosth previously here: https://github.com/OWASP/ASVS/issues/1612#issuecomment-2250605886. In addition, please see RFC 5764.

> 53.2.2 - "Verify that"? It seems general V9 requirement (not WebRTC specific)

We can update the text to state "Verify that". Also see my previous reply to @tghosth.

> 53.2.3 - publically known vulnerable sounds like a duplicate of 14.2.1
> 
> #	Description	L1	L2	L3	CWE
> **14.2.1**	Verify that all components are up to date, preferably using a dependency checker during build or compile time.	✓	✓	✓	1026
> race condition test is covered by 11.1.6. Is there something WebRTC-specific?
> 
> #	Description	L1	L2	L3	CWE
> **11.1.6**	[MODIFIED] Verify that the application uses synchronization and locking mechanisms for sensitive operations in order to keep internal data consistent, maintain user state, and prevent race conditions, such as 'time of check to time of use (TOCTOU)' vulnerabilities.		✓	✓	367

53.2.3 is not about a vulnerability that is necessarily product specific, which is what 14.2.1 is about. We describe the vulnerability itself, which is likely to be found in custom media servers or media server products that have not issued an advisory about it. Thus dependency checking or similar approaches are not as effective for this particular issue.

With regards to 53.2.3, yes this is very specific to WebRTC. If one were to look at 11.1.6, my understanding is that they would not be specifically looking for this vulnerability (which we document [here](https://www.enablesecurity.com/blog/novel-dos-vulnerability-affecting-webrtc-media-servers/)).

> 53.2.4, 53.2.5 - SRTP, RTP acronyms

I'll update all acronyms. 

> 53.2.5, 53.2.6 - slash (/) instead of "and" or "or"

Makes sense. 

> 53.2.6, 53.2.7, 53.3.1, 53.3.2 - "can handle" is a bit vague. What it means?

I'll define "can handle" in each case.

> I only checked requirement texts (not checked chapter texts).

Thanks for the work!

elarlang:
By principle, each requirement must be understandable as an independent line "out of context" - without including any other requirement, chapter texts, or glossary.

If it is not a well-known acronym, we need to write it. The "Well known" we can decide - if it is written in https://ahdictionary.com/word/search.html, it is well known. We don't have it formally documented (yet), but it is discussed here https://github.com/OWASP/ASVS/issues/1875

With "and/or" I expressed myself incorrectly - my goal is to avoid "/" instead of "and" or "or". So in this case I think instead of "video/audio" we can use "video or audio".

Thank you for this contribution, @sandrogauci !

sandrogauci:
Thanks for the explanation. Opened a new PR: #2023 to define the acronyms and change the slash into an "or".

> All "should" lines - are those recommendations (nice to have) or requirements (must have)?

I suppose it depends on your threat model. From where we stand, they are requirements since those are usually the effective solutions to the security issues that we're trying to prevent.

tghosth:
Hi @elarlang, can you confirm if the updates resolve all the points you raised?

@sandrogauci, thanks for your patience with this process, it seems like we are nearly there :)

sandrogauci:
> Hi @elarlang, can you confirm if the updates resolve all the points you raised?
> 
> @sandrogauci, thanks for your patience with this process, it seems like we are nearly there :)

No worries, it has been a learning process for me too! Thanks

tghosth:
Ok @elarlang I think you need to approve the changes now and then we can get this merged in :)

PR #2015, opened 2024-08-12
Update 0x50-V50-Web-Frontend-Security.md

jmanico:
Addressing issue https://github.com/OWASP/ASVS/issues/1965

PR #2020, opened 2024-08-14
redirect_uri validation for AS, resolves #1965

### LINKED ISSUE #1965: proposal/discussion: OAuth - separate requirement for redirect_uri string-match registration and handling
spin-off from https://github.com/OWASP/ASVS/issues/1916 "Discussion/Proposal 3"

Probably I was the first one to say that `redirect_uri` validation is a duplicate of general open-redirect but now I think it's important to have them as a separate requirement:

`redirect_uri` must be validated with the _string-match_ method, which means no "wildcard" validations.

There are 2 parts:

  * Authorization Server must not accept anything else than (one of) the precise URL from the pre-registered list
  * As a precondition: the OAuth Client must not send business logic URL to the Authorization Server. It is pretty much the same as Referrer leakage.

--
Feedback from @tghosth in https://github.com/OWASP/ASVS/issues/1916#issuecomment-2024985876

> Suggest you propose an updated/added requirement.

elarlang:
This Pull Request relates to issue #1965


tghosth:
I will approve this as the wording is what @jmanico wrote for #2015

tghosth (approved):

Issue #1990, opened 2024-07-10
Link checker is temperamental and apparently deprecated


tghosth:
The link checker seems to sometimes fail, even when it should not.

For example:
https://github.com/OWASP/ASVS/actions/runs/9854671316/job/27207915942

The link seems to work even through the link checker said it didn't.

Also, it says in the README that it is deprecated.
https://github.com/gaurav-nelson/github-action-markdown-link-check?tab=readme-ov-file

@ike do you think you could upgrade to the newer tool he suggests?

ike:
Yeah! Looks like he's also created [a new github action](https://github.com/UmbrellaDocs/action-linkspector).

tghosth:
Ok so do you want to try and upgrade it?

tghosth:
Hi @ike, any update on this?

arkid15r:
> It would also be good to only include 5.0 in the link checker and not 4.0 because we don't change 4.0 any more but we do add translations so it makes a mess: https://github.com/OWASP/ASVS/actions/runs/10573577586/job/29345695930?pr=1973
> 
> @ike @arkid15r

Alright, I'll work on migrating the workflow to a proper replacement and running it for `5.0` only. 
I'm going to look around first as https://github.com/UmbrellaDocs/action-linkspector looks pretty young action

> ERROR: 2 dead links found!
> [✖] https://owasp.org/www-project-security-knowledge-framework/ → Status: 404
> [✖] https://cheatsheetseries.owasp.org/cheatsheets/IndexASVS.html → Status: 404

The findings from https://github.com/OWASP/ASVS/actions/runs/10573577586/job/29345695930?pr=1973 look legit to me as they are 404. My understanding is they just irrelevant as `4.0` doesn't require any updates @tghosth ?

tghosth:
> My understanding is they just irrelevant as 4.0 doesn't require any updates

Correct 

> Alright, I'll work on migrating the workflow to a proper replacement and running it for 5.0 only.

Great!

arkid15r:
Just a small update:

As https://github.com/OWASP/ASVS/pull/2035 has been merged please let me know if you notice something is broken or works not as expected. I'll keep an eye on the workflow too.

The only part left to resolve this issue is addressing URL checker action deprecated status. I'm going to work on it this week.

tghosth:
Thanks so much!

tghosth:
Hi @arkid15r did you ever get to this?

arkid15r:
[...]

Issue #1900, opened 2024-03-17
7.2.3 (v4.0.3-7.1.3) - wording improvement


elarlang:
| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **7.2.3** | [MODIFIED, MOVED FROM 7.1.3] Verify that the application logs security relevant events including deserialization failures, input validation failures and incorrect HTTP requests (including requests with an unexpected HTTP verb). | | ✓ | ✓ | 778 |

Proposal: improve the wording to cover all decisions, which are made by the application to not accept user input.

Problems to solve:

  * "input validation failures" - this sounds more like a situation when the program code fails when validating user input, but should be the event when user input gets rejected by the application
  * to be sure it covers file upload situation (#1622)
  * to be sure it covers situations if JWT was not accepted

elarlang:
> Verify that the application logs attempts to bypass security controls.

This part ("security controls") is a bit maybe too abstract. The goal for this requirement should be to address input validation (as unexpected structure for the input), but in current wording it also covers access control and anything else.

> ... which are defined at both the client-side and a trusted service layer such as file upload controls or input validation.

The second part of the proposal is confusing. The message itself is (probably) good - you need to validate the input on the client side and then again on the server side

With current wording, it sounds like it is limited to detect only situations, where client-side validation exists, but it is bypassed. Is this the goal?

tghosth:
| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **7.2.3** | [MODIFIED, MOVED FROM 7.1.3] Verify that the application logs attempts to bypass security controls as defined in the relevant design documentation. | | ✓ | ✓ | 778 |

 We should make sure that the new [1.7.3](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x10-V1-Architecture.md#v17-errors-logging-and-auditing-architecture) sufficiently covers this.

~~which are defined at both the client-side and a trusted service layer such as file upload controls or input validation.~~

Add some intro text about the idea of the section and trying to abstract out. Don't eliminate 7.2.1 and 7.2.2 immediately.

tghosth:
Created #1944 to resolve this.

Additional events added to the logging cheatsheet here: https://github.com/OWASP/CheatSheetSeries/pull/1394

elarlang:
[...]

PR #2030, opened 2024-08-27
Fix URL redirect

arkid15r:
Related to https://github.com/OWASP/ASVS/issues/1990

It seems the link checker is not temperamental but is just doing its work. The problematic link doesn't return expected `200` but instead is a redirect. I tested it in my fork and [it worked](https://github.com/arkid15r/owasp-asvs/actions/runs/10581325636/job/29318321873).

As for the rest of the issue -- I think it's worth researching available markdown link check solutions as the current one is deprecated. My understanding it's just a wrapper around [markdown-link-check package](https://github.com/gaurav-nelson/github-action-markdown-link-check/blob/5c5dfc0ac2e225883c0e5f03a85311ec2830d368/entrypoint.sh#L11C10-L11C29).

It'd be a good staring point to decide whether redirects should be treated as accepted status codes.

I can look into addressing the rest of the #1990 if you'd like.


tghosth (approved):

tghosth:
Happy for you to take a look at #1900 @arkid15r but please write in that issue what you are planning on doing

PR #2035, opened 2024-08-30
Update URL Checker workflow

arkid15r:
This Pull Request relates to issue #1990 
This PR **does not** address the deprecated action update.

Update URL Checker workflow
  - merge 2 URL checker workflows (push, pr triggers)
  - limit runs to 5.0/** paths
  - bump actions/checkout from v3 to v4


.github/workflows/url-checker.yml
```diff
@@ -1,23 +1,41 @@
...
-          fetch-depth: 1
-      - name: link-check
+        uses: actions/checkout@v4
+
+      - name: Set options
+        id: options
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            echo "check-modified-files-only=yes" >> $GITHUB_OUTPUT
+            echo "use-quiet-mode=yes" >> $GITHUB_OUTPUT
```
tghosth, line 26: Would it be clearer to put this outside the if statement if the answer is the same either way or even move it back to being hard coded?

.github/workflows/url-checker.yml
```diff
@@ -1,23 +1,41 @@
...
+      - name: Check URLs
         uses: gaurav-nelson/github-action-markdown-link-check@v1
         with:
-          use-quiet-mode: 'yes'
-          use-verbose-mode: 'no'
-          config-file: '.github/workflows/config/url-checker-config.json'
-          folder-path: './5.0/'
+          check-modified-files-only: ${{ steps.options.outputs.check-modified-files-only }}
+          config-file: .github/workflows/config/url-checker-config.json
+          folder-path: '5.0'
```
tghosth, line 39: Does this change to the folder path make sure it only looks at links within 5.0?

.github/workflows/url-checker.yml
```diff
@@ -1,23 +1,41 @@
...
+  workflow_dispatch:
 
 jobs:
   markdown-link-check:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
-        with:
-          fetch-depth: 1
```
tghosth, line 16: Do we no longer need fetch depth?

tghosth (changes requested):

.github/workflows/url-checker.yml
arkid15r: That's correct, `1` is the [default value](https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4).

.github/workflows/url-checker.yml
arkid15r: Good point, done.

.github/workflows/url-checker.yml
arkid15r: I've modified the `on` part with `paths` options to make sure it runs only when '5.0/**' files are affected.

The triggers also include `workflow_dispatch` and `folder-path: '5.0'` should take care of those cases.

arkid15r (commented):
@tghosth thanks for reviewing:

tghosth:
Hi @arkid15r, thanks for your responses but I don't see that you actually made the changes to the PR?

arkid15r:
> Hi @arkid15r, thanks for your responses but I don't see that you actually made the changes to the branch in the PR?

Sorry for confusion. I forgot to update the PR yesterday.

tghosth (approved):

PR #2054, opened 2024-09-02
Merge in V5 refresh

tghosth:
After all associated issues are completed.


tghosth:
@elarlang would appreciate your final review on this. I don't think V5 is complete yet but this handles a lot of the issues and will cut down on the multiple open branches

elarlang (approved):

PR #2081, opened 2024-09-15
Fix use of allowlist throughout

tghosth:


tghosth:
We use the phrase allowlist in multiple places as either allow-list or allow list.

Having noticed this, I looked at the Bishop Fox style guide which as far as I know is the most comprehensive effort to standardize security terminology.

It notes the following:

image

Two key points:

- No space between allow and list
- Avoid using as a verb

This PR aligns ASVS with this.

elarlang (approved):

Issue #1903, opened 2024-03-17
move section V5.5 (and V5.4) to V10


elarlang:
The topic was raised by Fabio Cerullo in Slack chat.

Current section [V5.5 Deserialization Prevention](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x13-V5-Validation-Sanitization-Encoding.md#v55-deserialization-prevention) contains requirements which should be moved to "code review" category or section, we have V10 and V10.4 for those now.

The same applies to V5.4.

tghosth:
I don't agree, I stand by my previous comments [in the thread](https://owasp.slack.com/archives/C06MNF14M/p1709595823045979):

> _deserialisation vulnerabilities are a result of not handling dangerous input correctly which is why they're there (**in 5.x**)_ 
>
> _ultimately this is a problem of bad input which is why I think v5 is the closest compared to other chapters_
>
>  I would add that I think we would rather have related requirements together rather than being too strict about which chapter they belong to

jmanico:
I agree, Josh. But that is, of course, 1/2 that story.

Deserialization vulnerabilities occur when an application deserializes untrusted data without adequate validation or sanitization. An attacker can manipulate the serialized data to achieve arbitrary code execution, privilege escalation, or denial of service attacks.

While it is true that these vulnerabilities come from improperly handling input (which is often external), it's not just about "bad input" but also about how the application processes and uses this input. :D

elarlang:
Just for brainstorm.

I think we can take a parallel here with the so-called XSS - input validation MAY give you a defensive effect, but the application needs to guarantee that it can process the data securely.

**Example 1**:

  * _V5.5.2 Verify that the application correctly restricts XML parsers to only use the most restrictive configuration possible and to ensure that unsafe features such as resolving external entities are disabled to prevent XML eXternal Entity (XXE) attacks._

You may validate the incoming input, but the application is still vulnerable when meeting malicious input. The requirement does not say anything about input validation, but strict configuration to avoid XXE to happen even when not validated input reach to the functionality. The requirement is correct, but the choice of section is questionable.

The input validation part for processing XML is covered by other requirement:

  * _V13.3.1 Verify that XSD schema validation takes place to ensure a properly formed XML document, followed by validation of each input field before any processing of that data takes place._


**Example 2**:

[...]

PR #2116, opened 2024-09-25
Final wording tweaks on V5

tghosth:
Part of V5 reorg

### DIFF
--- a/5.0/en/0x13-V5-Validation-Sanitization-Encoding.md
+++ b/5.0/en/0x13-V5-Validation-Sanitization-Encoding.md
@@ -2,7 +2,7 @@
 
 ## Control Objective
 
-The most common web application security weakness is using untrusted content in an unsafe context without first making it safe. This weakness leads to many of the significant vulnerabilities in web applications, such as Cross-Site Scripting (XSS), SQL injection, OS command injection, template injection, log injection, LDAP injection, and more.
+This chapter focuses on the most common web application security weaknesses that are related to unsafe processing of untrusted data. This leads to a variety of technical vulnerabilities where the untrusted data gets interpreted using the syntax rules of the relevant interpreter.
 
 With modern web applications, it will always be best to use safer APIs such as parameterized queries, auto-escaping templating frameworks, etc. Otherwise, or carefully performed output encoding/escaping or sanitization will be critical to the security of the application.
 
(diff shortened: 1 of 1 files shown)


elarlang (commented):
In general for each sentense there should be clear goal, why it exists

> The most common web application security weakness is using untrusted content in an unsafe context without first making it safe.

What the "first making it safe" means here?

> Note from Josh, I think this section should maybe be moved further down as it is less important, I would consider it for 5.4

Input validation - sanitization - output encoding is logical way hot application process the input and it is the same order for testing it. So I prefer the order to stay.


> Input can come from a variety of sources including HTML form fields, REST requests, URL parameters, HTTP headers, cookies, files on disk, databases, external APIs, etc.

Everything the application uses or processes must be taken as user input, including HTML form fields, REST requests, URL parameters, HTTP headers, cookies, files on disk, databases, external APIs, etc.


What is the safe API?

tghosth:
> In general for each sentense there should be clear goal, why it exists
> 
> > The most common web application security weakness is using untrusted content in an unsafe context without first making it safe.
> 
> What the "first making it safe" means here?

I mean, the whole chapter is about how to make it safe, this is just the first line which summarised the issue.

> > Note from Josh, I think this section should maybe be moved further down as it is less important, I would consider it for 5.4
> 
> Input validation - sanitization - output encoding is logical way hot application process the input and it is the same order for testing it. So I prefer the order to stay.

So I see it as order of preference for solution, best is output encoding, then sanitization, then input validation as defense in depth

> > Input can come from a variety of sources including HTML form fields, REST requests, URL parameters, HTTP headers, cookies, files on disk, databases, external APIs, etc.
> 
> Everything the application uses or processes must be taken as user input, including HTML form fields, REST requests, URL parameters, HTTP headers, cookies, files on disk, databases, external APIs, etc.
> 
> What is the safe API?

What is your question? Are you saying that pretty much anything is untrusted input? I guess known constants might be considered trusted input...

elarlang:
My point of view is: not making it safe, but process it safely. Making it safe sounds like sanitization.

>  I guess known constants might be considered trusted input...

Philosophical, but I think is one of the main sources of why vulnerabilities happen.

For example, for output encoding, it does not matter, from where the input came. If the "trusted constant" contains symbols that require converting or escaping, you need to do it anyway.

tghosth:
Can you suggest alternative wording

jmanico:
> So I see it as order of preference for solution, best is output encoding, then sanitization, then input validation as defense in depth

I agree with Elar’s philosophy: it's not about making input safe but about processing input safely.

However, I politely disagree with the idea of an order of preference for solutions. The best control depends on the specific context. For example, when dealing with HTML input, sanitization is the only correct way to handle it safely. For untrusted URLs, validation is often the most appropriate method. In other cases, parameterization might be the correct approach, or using a modern function for file I/O could ensure safety.

The right control must be used for the right situation. There isn’t a single order of preference; it’s about selecting the appropriate defense based on the context.

elarlang:
**Control Objective**

A bit of content-heavy concentrate, but for serving the idea:

_This chapter focuses on the most common web application security weaknesses that are related to unsafe processing of the data through the application from being technically correct - from missing or incorrect input validation if accepting the data to missing or incorrect way to encode, escape, or sanitize the data for the output or for the next component._

As the first line addresses 3 topics, but there are also V5.4 and V5.5 need to be addresed (although I have proposed to move those to V10 https://github.com/OWASP/ASVS/issues/1903)

**Input validation**

First line:
_Everything the application uses or processes must be handled as user input, including HTML form fields, REST requests, URL parameters, HTTP headers, cookies, files on disk, databases, external APIs, etc._


> Input validation provides valuable security hygiene and should be applied to all inputs where possible. However, since input validation is not able to prevent specific vulnerabilities, it should be used as a primary security defense. Other techniques detailed in later sections should be used as the primary defense against specific vulnerabilities.

This is quite far of being correct for me.

The message should be:
_The goal for input validation is to provide data for the application in the expected format (this part duplicates the previous line a bit and needs some work) and it does not remove or replace the need to use correct encoding, escaping, or sanitization when using the data for next component or for presenting it for the output._

Or, just remove the line.


**Output encoding**

> Ideally, safe APIs should be used which will do this automatically but otherwise the developer will need to take the required action.

What is the magic "safe API" that solves all the output problems? There is no such thing. As I don't understand the point or the message for the line, I can not provide any recommendations.

jmanico:
> What is the magic "safe API" that solves all the output problems? There is no such thing.

However, there are many examples of functions in various languages that address specific security concerns. For instance, file I/O functions in several languages can prevent path traversal attacks, and functions that enforce the use of parameterized queries (like Criteria-based queries) protect against SQL injection.

While I’m not claiming that any single "safe API" can solve all output problems—that would be an oversimplification of what I was trying to say—there are many functions that, when used correctly, significantly improve security. These functions generally offer a safer, more reliable approach than handling such issues manually.

jmanico:
Here is recent article that talks about how using safe functions in certain languages have a dramatic effect on security. This is the general concept that I'm talking about.

https://security.googleblog.com/2024/09/eliminating-memory-safety-vulnerabilities-Android.html

tghosth:
@elarlang please let me know if you want further changes

elarlang (approved):

jmanico (approved):

PR #2125, opened 2024-10-08
Clarify 5.2.6 ssrf requirement to resolve #2115

### LINKED ISSUE #2115: Deduplicate SSRF requirements
We currently have the following requirements which refer to SSRF:

| **5.2.6** | [GRAMMAR] Verify that the application protects against SSRF attacks, by validating or sanitizing untrusted data or HTTP file metadata, such as filenames and URL input fields, and uses an allowlist of protocols, domains, paths and ports. | ✓ | ✓ | ✓ | 918 |
| **5.5.5** | [MODIFIED, MOVED FROM 13.1.1, LEVEL L1 > L2] Verify that different parsers used in the application for the same data type (e.g. JSON parsers, XML parsers, URL parsers), perform parsing in a consistent way and use the same character encoding mechanism to avoid issues such as JSON Interoperability vulnerabilities or different URI or file parsing behavior being exploited in Remote File Inclusion (RFI) or Server-side Request Forgery (SSRF) attacks. | | ✓ | ✓ | 436 |
| **12.3.1** | [MODIFIED, MERGED FROM 12.3.2, 12.3.3, 5.3.9] Verify that file operations avoid using user-submitted filenames or file metadata when creating file paths to protect against path traversal, local or remote file inclusion (LFI, RFI), and server-side request forgery (SSRF) attacks. Instead, use internal, trusted data for file I/O. If user-submitted filenames or file metadata must be used, strict validation and sanitization must be applied. | ✓ | ✓ | ✓ | 73 |

I think that requirements 5.5.5 and 12.3.1 are referring to specific protection mechanisms which are required which have the potential side effect of SSRF as such, I think they are ok as they are.

I think that 5.2.6 needs to be very slightly clarified to focus on the risk of putting input directly into a URL being accessed. 

I would say:

| **5.2.6** | [MODIFIED] Verify that the application protects against SSRF attacks, by checking untrusted data against an allowlist of protocols, domains, paths and ports and sanitizing potentially dangerous characters before using the data to call another service. | ✓ | ✓ | ✓ | 918 |

Thoughts @jmanico @elarlang

tghosth:
This Pull Request relates to issue #2115


elarlang (approved):
approve, but should we write SSRF as server-side request forgery into the requirement?

tghosth:
We do define it in the glossary, do you think it is necessary?

elarlang:
> We do define it in the glossary, do you think it is necessary?

We may expect that our target audience knows what the SSRF is. If we take some guidelines on how to handle the situation, then if (for example) https://ahdictionary.com/word/search.html?q=SSRF does not know the acronym, then you should not use only the acronym.

tghosth:
> > We do define it in the glossary, do you think it is necessary?
> 
> We may expect that our target audience knows what the SSRF is. If we take some guidelines on how to handle the situation, then if (for example) https://ahdictionary.com/word/search.html?q=SSRF does not know the acronym, then you should not use only the acronym.

I am not sure I understand your answer, can you clarify please :)

elarlang:
```
if ('https://ahdictionary.com/word/search.html?q=' + acronym) == 404
	use 'full name (acronym)'
```

elarlang:
https://github.com/OWASP/ASVS/pull/2125/commits/0d73972adf072d88bacf65d30bb5070e5e67e267

I think your keyboard does not work properly today.

Forgery, SSRF, closing bracket...

PR #2200, opened 2024-10-28
Updated Glossary to Resolve #2100

### LINKED ISSUE #2100: V3 Terminology Addition
I think it is necessary to provide definitions of terms used within V3 (and V1.3) and correlate to related and competing definitions. This is probably best added to the glossary when time to reorganize, but I would like to confirm terms for other V3 work. I added comments as indented bullets below items. Starting with the following:

- **Stateful Mechanism or Token** - These mechanisms utilize the application server to retain session state and token data in full.

- **Stateless Mechanism or Token** - These mechanisms encode information within the token(s) passed to clients that is not maintained by the server. A server may store *some* session information (such as a JWT revocation list) and be considered a stateless design.
  - This definition settles ongoing debate over what constitutes a Stateless session by simply defining a version where we permit some server-side state.

- **Opaque Token** - An opaque token is a token with no discernible structure or meaning outside of the issuing server. 
  - Note: I have defined opaque above as it appears used across various industry players, but V3 appears to conflate opaque tokens (original use in #1478 (@tghosth)] with randomly generated stateful tokens. This use is not universal. For example, Auth0 defines them with "Tokens in a proprietary format typically contain some identifier to information in a server’s persistent storage" and I also came across [Ory](https://www.ory.sh/docs/oauth2-oidc/jwt-access-token) which uses a signature with "opaque" tokens. Some additional requirement rework may be appropriate for clarity.
  - I propose that a more clear and unconflicted term is used for this purpose (randomly generated stateful tokens). Please consider some candidate terms:  Stateful Token, Server-Side Token, Random Token, Identifier Token, or just Randomly Generated Stateful Tokens.
  - Otherwise, I would suggest we simply adopt our own definition of Opaque Token as a randomly generated, high entropy string that is stored server-side to track sessions.

- **Absolute Maximum Session Lifetime** - Also referred to as "Overall Timeout" by NIST, this is the maximal amount of time a session can remain active following authentication regardless of user interaction. This is a component of session expiration.

- **Inactivity Timeout** - This is the length of time a session can remain active in the absence of user interaction with the application. This is a component of session expiration.

- **Cryptographically Signed Token** - These tokens often contain claims or assertions, relying on cryptographic techniques such as signatures or message authentication codes (MAC) to ensure authenticity and integrity of data encoded within the token structure. 

Final note: it may be worth defining "Session ID" and "Session Token". These appear to be used interchangeably by documents such as [OWASP Cheat Sheets](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html). I think of "Session ID" as an identifier unique to a session whereas a "Session Token" is the actual string/data transmitted between client and server. They *could* be one in the same.

ryarmst:
This Pull Request relates to issue #2100


tghosth:
@elarlang please can you review my addition of JWT and SAML, I reviewed the rest.

I think the alphabetical order of the glossary is broken so I opened #2201 to check this when we are near the release

elarlang (approved):

Issue #959, opened 2021-04-07
New Requirement on Browser Version Checking


jmanico:
So much of ASVS is beginning to address new web standards like SameSite cookies.

I feel all websites should programmatically limit browsers in some way to ensure that very old browsers with bad web standard cannot be used my typical users.

Sure attackers can circumvent this but that is not the point.
 
For typical users (and for CSRf like attack scenarios) I feel JS block older browsers and to enforce a clear browser standard is crucial for the secure web.

elarlang:
Does it makes sense from ASVS point of view to say, that you can not use certain (not EOL) browser, even site has proper defense for used technologies in place?

Supporting EOL browser versions actually does not makes sense, because if those will not get security patches, site owner need to consider those clients as "hacked clients" - confidential information may leak and so on.

jmanico:
And older browsers do not support modern TLS, samesite, CORS, CSP 2 and a host of other standards that secure web apps need.

elarlang:
But in practice, how could it work? From where you'll get list of "valid browsers"?

Just based on User-Agent string alone:
* you can not use allow-list - other-wise you keep away some "weird" users. No one want to loose their visitors.
* you can not use deny-list - you just can not keep a list of all possible "not good enough" browsers

cmlh:
An example of where support has to be given to the widest older releases of web browsers is when the public have to browse emergency warnings, etc yet this web app is considered as "high assurance" i.e. Level 3

The solution is to highlight when each ASVS Requirement supports only the latest browser release similar to https://caniuse.com/

elarlang:
> An example of where support has to be given to the widest older releases of web browsers is when the public have to browse emergency warnings, etc yet this web app is considered as "high assurance" i.e. Level 3

This is interesting problem
* it does not make sense to hide information about emergency warnings, because it may be critical information
* it does not make sense to show this information because you can not trust the client (browser) and integrity of the messages is questionable

But before we go there, my question is still - is it technically doable (with reasonable effort) without having detailed user-agent information?

cmlh:
> But before we go there, my question is still - is it technically doable (with reasonable effort) without having detailed user-agent information?

Yes as the web application is relying upon the web browser technologies cited by https://github.com/OWASP/ASVS/issues/959#issuecomment-815155670 rather than `User-Agent` Request Header alone.

The caveat is if the web application itself doesn't implement these web technologies to deliver the web application then it is simply increasing its attack surface.

[...]

PR #2205, opened 2024-10-28
#959 - document requirements for browsers and check them

elarlang:
This Pull Request relates to issue #959

### DIFF
--- a/5.0/en/0x10-V1-Architecture.md
+++ b/5.0/en/0x10-V1-Architecture.md
@@ -155,6 +155,12 @@ This is a placeholder for future documentation requirements.
 | **1.14.6** | [MOVED TO 50.7.2] | | | | |
 | **1.14.7** | [MODIFIED, MOVED FROM 1.1.5] Verify that all communication needs for the application are documented. This should include external services which the application relies upon and cases where an end user might be able to provide an external location to which the application will then connect. | | ✓ | ✓ | 1059 |
 
+## V1.50 Web Frontend Security Documentation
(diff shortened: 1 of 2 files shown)


5.0/en/0x10-V1-Architecture.md
```diff
@@ -155,6 +155,12 @@ This is a placeholder for future documentation requirements.
 | **1.14.6** | [MOVED TO 50.7.2] | | | | |
 | **1.14.7** | [MODIFIED, MOVED FROM 1.1.5] Verify that all communication needs for the application are documented. This should include external services which the application relies upon and cases where an end user might be able to provide an external location to which the application will then connect. | | ✓ | ✓ | 1059 |
 
+## V1.50 Web Frontend Security Documentation
+
+| # | Description | L1 | L2 | L3 | CWE |
+| :---: | :--- | :---: | :---: | :---: | :---: |
+| **1.50.1** | [ADDED] Verify that application documentation states the expected security features that browsers using the application should support (such as HTTPS, HSTS, Content Security Policy (CSP), and other relevant HTTP security mechanisms). It should also define how the application must behave when such an event happens (such as warning the user or blocking access). | | | ✓ | |
```
randomstuff, line 162: Actually "such an event happens" doesn't work here (because we did not tell which event we are talking about). This should be something such as:

> It should also define how the application must behave when some of these features are not available (such as warning the user or blocking access)

tghosth (approved):

PR #2211, opened 2024-10-31
Clear up sandboxing requirements to resolve #2166

### LINKED ISSUE #2166: Review requirementes 14.2.6 and 14.2.8, potential move to V10
Spin-off from #2088 / the discussion over 14.2.6 and/or 14.2.8 comes from #1425.

Current requirements:

| **14.2.6** | [MODIFIED, SPLIT TO 14.2.8, LEVEL L2 > L3] Verify that risky third party libraries or those with a history of vulnerabilities are encapsulated such that only required behaviour is available to the application, to reduce attack surface. | | | ✓ | 1061 |
| **14.2.8** | [ADDED, SPLIT FROM 14.2.6] Verify that risky third party libraries or those with a history of vulnerabilities are sandboxed away from the most sensitive system modules/services so that even if a vulnerability in the library was successfully exploited, the sensitive system modules/services would not be compromised. | | | ✓ | 1061 |

The need for those requirements are questioned for example in comments:
* https://github.com/OWASP/ASVS/issues/2088#issuecomment-2355412099
* https://github.com/OWASP/ASVS/issues/1425#issuecomment-1745209042

For me the "main error" is, if we talk about L3 requirement, then those application can not use any "too risky" and not trustful component anyway.

By content, both are more "software architecture" requirements, not a clear configuration requirements. Most likely we need a chapter for V10 for that, something to say "Software Architecture" or "Sandboxing".

tghosth:
This Pull Request relates to issue #2166

### DIFF
--- a/5.0/en/0x10-V1-Architecture.md
+++ b/5.0/en/0x10-V1-Architecture.md
@@ -121,8 +121,8 @@ Architectural requirements are intrinsic to the entire code base, and thus diffi
 | :---: | :--- | :---: | :---: | :---: | :---: |
 | **1.10.1** | [DELETED, NOT IN SCOPE] | | | | |
 | **1.10.2** | [MODIFIED, MOVED FROM 14.2.5, MERGED FROM 14.2.4] Verify that an inventory catalog, such as software bill of materials (SBOM), is maintained of all third-party libraries in use, including verifying that components come from pre-defined, trusted, and continually maintained repositories. | | ✓ | ✓ | |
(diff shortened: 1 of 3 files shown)


elarlang:
I changed tags:
```
1.10.3 [ADDED, SPLIT FROM 14.2.6]
1.10.4 [ADDED, SPLIT FROM 1.14.5]
1.14.5 [SPLIT TO 1.10.4, 10.5.1]
10.5.1 [ADDED, SPLIT FROM 1.14.5, 14.2.6]
14.2.6 [SPLIT TO 1.10.3, 10.5.1]
```

We can do one way and another way and none of them seemed correct, but the logic for changes:
* when we split one requirement to many, we don't delete the meaning of the requirement from ASVS but spread it to 2 new requirements
* when it is split, the other requirement is ADDED
* for ADDED we watch it as new requirement and don't carry LEVEL information from previous one.

tghosth:
Fair one, if you are happy with the content then please approve

elarlang (approved):

PR #2236, opened 2024-11-05
tag fix for #2223

elarlang:
removed requirement:
V12.3.6 Verify that the application does not include and execute functionality from untrusted sources, such as unverified content distribution networks, JavaScript libraries, node npm libraries, or server-side DLLs. | | ✓ | ✓ | 829 |

With duplicate of we should point to v4.0.3 requirements:
* V14.2.3 Verify that if application assets, such as JavaScript libraries, CSS or web fonts, are hosted externally on a Content Delivery Network (CDN) or external provider, Subresource Integrity (SRI) is used to validate the integrity of the asset.
* V14.2.4 Verify that third party components come from pre-defined, trusted and continually maintained repositories.

This Pull Request relates to issue #2223


tghosth:
Ah ok, I see your point, thanks :)

tghosth (approved):

Issue #1544, opened 2023-02-17
Expand 1.6.3


jmanico:
With the upcoming quantum computing horizon coming up (2030) i think it's REALLY important for developers to achieve "cryptographic agility". So perhaps augment the current 1.6.3:

1.6.3 | Verify that all keys and passwords are replaceable and are part of a well-defined process to re-encrypt sensitive data. |   | ✓ | ✓ | 320
-- | -- | -- | -- | -- | --

...to include even more...

1.6.3 | Verify that all keys, passwords, cryptographic algorithms, key/secret vaults and other cryptographic components are replaceable and are part of a well-defined process to re-encrypt sensitive data. |   | ✓ | ✓ | 320
-- | -- | -- | -- | -- | --

#cryptographicAgility

tghosth:
We talk about this elsewhere in the standard. I agree with the concept but we need to investigate whether we need to do deduplication. I'll add this to my todo list...

danielcuthbert:
Added in 5.0 under V6.9 Post-Quantum Cryptography (PQC) and more specifically

6.9.2 	[ADDED] Verify that cryptographic systems are designed to allow for seamless upgrades to post-quantum cryptography, enabling the transition once PQC standards are fully established.

I don't feel we need this in 1.6.3 too so my approach would be to keep it as 6.9.2 and remove it from 1.6.3

PR #2272, opened 2024-11-05
moved 1.6.3 to 6.9.2

danielcuthbert:
This Pull Request relates to issue #1544
Moved this from V1 to V6


jmanico (approved):

elarlang (commented):
Removable:
* V1.6.3 Verify that all keys and passwords are replaceable and are part of a well-defined process to re-encrypt sensitive data.

Referenced 6.9.2:
* V6.9.2 Verify that cryptographic systems are designed to allow for seamless upgrades to post-quantum cryptography, enabling the transition once PQC standards are fully established.

Is that connection here correct?

tghosth:
@danielcuthbert is 1.6.3 basically requiring crypto agility? I would have thought that crypto agility is a more generic point than just for PQC, i.e. in general it should be a property of any good crypto system. Do we have a more generic requirement on that in V6?

tghosth:
Discussed with Daniel, I will merge 1.6.3 and 6.9.2 into 6.2.4 ensuring that it is worded in a modern way, mentions crypto agility and also the move to PQC.

Issue #1974, opened 2024-05-27
4.3.1 and 4.3.3


EnigmaRosa:
Addressing #1352 Q7:

> Q7 - 4.3.1, 4.3.3 - are those actually authentication requirements?

While 4.3.1 is relevant to access control - as access to the admin interface can enable changes to user permissions - I would argue that this is an authentication issue, not an access control one. Furthermore, the language surrounding "administrative interfaces" should be clarified, as I've seen users interpret it as admin interface for a single corporate instance (in a multi-tenant SaaS app).

I'm also inclined to argue that 4.3.3 doesn't belong in V4, and likely in authentication.

| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **4.3.1** | [MODIFIED] Verify administrative interfaces can only be logically accessed from trusted endpoints or locations. For example, restricting access to bastion or jump hosts, trusted admin workstations or endpoints (e.g., device authentication), administrative LANs, etc. | ✓ | ✓ | ✓ | 419 |
| **4.3.3** | [MODIFIED] Verify that, if the application allows changing highly sensitive configurations around passwords or connection parameters for integrations with databases and third-party systems, they are protected by extra controls such as re-authentication or multi-user approval. | | ✓ | ✓ | 732 |

elarlang:
> Q7 - 4.3.1, 4.3.3 - are those actually authentication requirements?

So the quote is from me. When I watch updated requirements, I don't think the authentication is suitable for those, as the responsibility or the goal is not identifying the user.

Some thoughts:
* 4.3.1 - as it is not under "application permission" logic and is/can be external, outside of application solution, we can watch it also as some kind of configuration
* 4.3.3 - it's authorization + logic for specific case, in a way related discussion https://github.com/OWASP/ASVS/issues/1576
  * re-authentication in the requirement is used as a tool for authorization, but the goal for the requirement is not related about procedure of identifying the user

In summary: if we don't find a clear better place, I'm ok when they stay where they are.

EnigmaRosa:
I understand. For 4.3.1, placing in configuration might be the best option, as well as having clarifying language. Is Business Logic perhaps a more suitable location for 4.3.3?

elarlang:
Improving the wording is always an option, but those requirements have seen some journeys already :)

As both requirement were modified after the last release, it makes sense to review the reasons how those reached to the current wording
* 4.3.1
[...]

PR #2296, opened 2024-11-06
Resolve #1576 by adding dual approval for very sensitive operations

### LINKED ISSUE #1576: Mitigate Risk of Fraudulent Activities by Implementing Dual Authorization [Business Logic Issue]
One of the most critical security issues that can exist in high-value sensitive systems (for example, banking systems) is the lack of dual authorization for sensitive operations or transactions. This issue occurs when the application allows a single user to perform a sensitive action or transaction or action without requiring approval from another authorized user.

For example, suppose a bank's online banking application allows a single bank clerk to transfer a large amount of money to another account without requiring approval from another authorized user. In that case, this creates a significant risk of fraudulent activity. We can have a malicious insider that can perform the transaction without the need for a second person to approve it, potentially resulting in a significant loss of funds for the bank and its customers.

To prevent such scenarios, it is essential to implement dual authorization for sensitive transactions or operations in high-value sensitive systems. This means that the application should require approval from at least two authorized users before processing the transaction. 

I would suggest this check:
- Verify that the application implements dual authorization for sensitive operations or transactions, requiring independent approval from multiple authorized users before processing.

jmanico:
This is not a clear access control requirement per https://github.com/OWASP/ASVS/issues/1974


jmanico (commented):
I am pushing a bit too hard on this and am ok to back it out @elarlang

tghosth:
I think this is actually more related to #1576 than 4.3.3

jmanico:
I like this version of 11.1.10

PR #2299, opened 2024-11-06
Move authentication with external services to 14.7 to resolve #2298

### LINKED ISSUE #2298: Creating a section in configuration for External Service Configuration
We have requirements (2.10.x and 4.3.3) which seem like AuthN or access control really they come down to how the app interacts with external service.

I think it's therefore very important to move these into a dedicated section in the configuration chapter for accuracy and because I think they are causing confusion and distraction in their current chapters.

tghosth:
This Pull Request relates to issue #2298


jmanico (approved):
my approval relates to 4.3.3

PR #2365, opened 2024-11-10
[IN PROGRESS] Improve front text

tghosth:


tghosth:
@elarlang can we get this work in progress merged in so we don't end up out of sync?

elarlang (approved):

PR #2386, opened 2024-11-19
Clarify 6.6.5

tghosth:
This is a wording tweak so I didn't create an issue. I wanted to make the requirement slightly clearer.

@danielcuthbert do you think you could add any detail about what would give more info about what a suitable algo would be.

Otherwise, if it is ok then please explicitly "Approve" from the "Files Changed" tab above to the right.


danielcuthbert (approved):
LGTM

PR #2387, opened 2024-11-19
Clarify 6.7.1

tghosth:
Didn't open an issue as this is just a wording tweak. I think I maintained the spirit of the requirement but please let me know @danielcuthbert . 

If it is ok then please explicitly "Approve" from the "Files Changed" tab above to the right.


danielcuthbert (approved):
This reads well, thank you @tghosth

PR #2389, opened 2024-11-20
Updates to section 1.6

### LINKED ISSUE #685: 1.6.4 requirement is too complicated to understand
1.6.4  Verify that symmetric keys,   passwords, or API secrets generated by or shared with clients are used only   in protecting low risk secrets, such as encrypting local storage, or   temporary ephemeral uses such as parameter obfuscation. Sharing secrets with   clients is clear-text equivalent and architecturally should be treated as   such.

I'm not a native english speaker and it's completely not clear for me what does it mean:
- "are used only   in protecting low risk secrets" - symmetric keys, passwords and so on are used to protect low risk secrets ( such as ..)? and what about high risk secrets? 
OR the meaning is "are used in protected low risk way (such as ...)";

- "temporary ephemeral uses such as parameter obfuscation" - temporary = parameter obfuscation?  OR parameter obfuscation should be temporary measure? 

"Sharing secrets with   clients is clear-text equivalent and architecturally should be treated as   such." - what is "as such"?

**why not to form it smth like

Verify that symmetric keys, passwords, API secrets, keystorages, certificates and other secrets shared with clients are used in protected low risk way (for example, they are stored in encrypted storage on the client side, they are temporary, they are obfuscated)..**

tghosth:
This Pull Request relates to issue #685


danielcuthbert (approved):
Amazing! having this in makes me happy. 100% approved

PR #2391, opened 2024-11-20
Try and slightly clarify scope of chapters to resolve #2375

### LINKED ISSUE #2375: V6 - "Stored cryptography" vs "cryptography"
The V6 chapter is now called "Stored Cryptography" (and was before named "Cryptography"). I actually missed that point until now.

Do we want V6 to be only about Stored Cryptography only?

Some things here would me applicable in other context (such as "Random Values").

If we want V6 to be about "Stored Cryptography", this should probably be clearer in the introduction. In this case, #2310 for example should go somewhere else.

If we keep V6 "Stored Encryption" and V9 "Communication Encryption":

* What about general cryptographic requirements which applies to both? (such as "Random Values")
* Is V9 about "transport layer" security only or should application-layer crypto be in V9 as well?

Related #2252, #2310

tghosth:
This Pull Request relates to issue #2375

### DIFF
--- a/5.0/en/0x14-V6-Cryptography.md
+++ b/5.0/en/0x14-V6-Cryptography.md
@@ -2,7 +2,7 @@
 
 ## Control Objective
 
-The objective of V6 is not only to define best practices but also to instill a fundamental understanding of cryptographic principles and inspire a shift toward more resilient and modern approaches by doing the following:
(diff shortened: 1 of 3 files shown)


danielcuthbert (approved):
breaking it down like this works incredibly well. 100% approve

PR #2432, opened 2024-12-05
Add no redirect to HTTPS to resolve #2416

### LINKED ISSUE #2416: Don't redirect to HTTPS for API
[Your API Shouldn't Redirect HTTP to HTTPS](https://jviide.iki.fi/http-redirects)

The argument is that when a client uses 'http://api.example.org', it should fail instead of silently be insecure.

I propose to add a requirement that specifies what behavior a site should have when the HTTP version is accessed, whether it should redirect or return an error.

tghosth:
This Pull Request relates to issue #2416


tghosth:
Your wording leads to a negative requirement. I agree with the level comment and I also clarified the second part of the req.

I would propose:

| **13.1.8** | [ADDED] Verify that HTTPS-based endpoints will only respond to non-encrypted HTTP requests with an error or will not respond at all. Responding with an automatic redirect to the HTTPS endpoint may lead to clients accidentally sending data over non-encrypted HTTP, but this is not being discovered.
 | | | ✓ | |

elarlang:
But I would still say level 2. It's easy to implement.

tghosth:
Fair comment @elarlang, I updated the PR so are you comfortable to approve now?

jmanico:
I know this is merged but this text is a little awkward.

>  non-encrypted HTTP 

This is redundant.

> requests with an error

This alone can be problematic. The better call it to not respond at all:

So I suggest we just keep this simple, with text like.

Verify that HTTPS-based endpoints do not respond to HTTP requests.

This drops the negative requirement of not redirecting, drops the "error message" part, and simplifies the language.

elarlang:
You can re-open related issue (#2416) with recommendations.

PR #2434, opened 2024-12-05
Clarify token terminology throughout to resolve #2414

### LINKED ISSUE #2414: Terminology and glossary update for token clarity
Following from #2384, here is an initial proposal for some changes:

### New Terms

**Self-Contained Token** - A token that encapsulates one or more attributes that do not rely on server-side state or other external storage. These tokens ensure the authenticity and integrity of their contained attributes, enabling secure, "stateless" information exchange across systems.

@randomstuff Should we define **Reference Token**?

### Updated

**Cryptographically Secured Token** -  A token protected using cryptographic techniques, such as digital signatures or message authentication codes (MACs), to ensure the authenticity and integrity of its data. These tokens can range from simple signed identifiers to more complex formats like JSON Web Tokens (JWTs).

tghosth:
This Pull Request relates to issue #2414

@ryarmst @elarlang @randomstuff @TobiasAhnoff


elarlang (changes requested):
I feel we are looping with this one and PR is different than agreed in the related issue.

* I prefer V52 title to be "Self-contained tokens" as it was.
* V4.2.5 should use self-contained tokens
* "Self-Contained Token - A subset of cryptographically secured tokens" - I really don't like that. It's not a subset, self-contained tokens are using cryptography for integrity. Whatever uses crypto for integrity is not subset of it, those are just using it. I think those proposals were good: https://github.com/OWASP/ASVS/issues/2414#issuecomment-2506815656

tghosth:
Commented in the issue

5.0/en/0x12-V3-Session-management.md
```diff
@@ -108,7 +108,7 @@ Session termination may be handled either by the application itself or by the SS
 
 Session termination should result in requiring re-authentication and be effective across the application, federated login (if present), and any relying parties.
 
-For stateful session mechanisms, this should just require invalidating the session at the backend. If a stateless session mechanism with signed tokens is being used, a way to revoke these tokens will be needed.
+For stateful session mechanisms, this should just require invalidating the session at the backend. If self-contained tokens are being used, a way to revoke these tokens will be needed.
```
elarlang, line 111: revoke or block

elarlang (approved):
Let's get those in. If needed, we can do another update.

PR #2441, opened 2024-12-06
rm reference

elarlang:
@danielcuthbert - I think this `**` was initially meant to be reference to the table below the requirement section. As those tables are moved away anyway, then this `**` is not valid anymore (and should not be used anyway, because requirement must be independent and without references)


danielcuthbert (commented):
Happy with this and yeah if the original wasn't to spec, apologies.

elarlang:
> Happy with this and yeah if the original wasn't to spec, apologies.

please approve the PR then :)

tghosth (approved):

Issue #2468, opened 2024-12-15
clarify 50.6.2


elarlang:
| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **50.6.2** | [ADDED, SPLIT FROM 5.3.3] Verify that JavaScript context-aware methods are used when handling untrusted data to avoid unintended content execution, such as executing content as HTML instead of displaying it as text. | ✓ | ✓ | ✓ | |

Seems that this requirement needs clarification or an example.

Something like:
> For example, untrusted input must not be applied via innerHTML or outerHTML properties, or not used via setHTMLUnsafe, insertAdjacentHTML or document.write functions. Instead, createTextNode should be used.

elarlang:
ping @randomstuff - I saw "in some other source" that this requirement was confusing for you, does the proposed addon make sense and make it more clear?

randomstuff:
Thanks, I intended to make a pass of the requirements I did not understand at some point and either try to understand them or raise an issue :)

The thing I did not understand was "context-aware methods" but with the examples I understand what we are talking about. So that's a clear improvement.

(However, maybe we could make this clearer or make explicit that (AFAIU) interpretation of the user-provided HTML may depend on where we are on the DOM. Searching "context-aware JavaScript" does not give much help. I don't have any suggestion for now, however.)

jmanico:
> For example, untrusted input must not be applied via innerHTML or outerHTML properties, or not used via setHTMLUnsafe, insertAdjacentHTML or document.write functions. Instead, createTextNode should be used

As a side note, here are functions in JS that are safe besides createTextNode

- elem.textContent = dangerVariable;
- elem.insertAdjacentText(dangerVariable);
- elem.className = dangerVariable;
- elem.setAttribute(safeAttribute, dangerVariable); 
- formfield.value = dangerVariable;
- document.createTextNode(dangerVariable);
- document.createElement(dangerVariable);

And anything else can be made safe with DOMPurify (which is also an experimental native browser function)

- elem.innerHTML =DOMPurify.sanitize(dangerVariable)l

elarlang:
I think we just need to pick one most widespread function as an example to explain the idea behind the requirement.

I would not recommend DOMPurify here as this does not display the input **as is** - it changes the value and is part of sanitization.

jmanico:
> I think we just need to pick one most widespread function as an example to explain the idea behind the requirement.

[...]

PR #2479, opened 2024-12-17
#2468 no innerHTML and document.write

elarlang:
This Pull Request relates to issue #2468


jmanico (approved):

randomstuff:
> [ADDED, SPLIT FROM 5.3.3] Verify that untrusted input must not be applied via innerHTML, document.write, or other properties or functions that render HTML. Instead, use createTextNode, textContent, and similar safe functions that do not render HTML and only render content as text.

What about when the user enters markdown or HTML (which is then filtered using some library). Does "filtered user input" count as "untrusted input"?

I would consider it is not allowed by this wording requirement but should be allowed. Should we add some provision for when the input has been properly filtered?

elarlang:
> > [ADDED, SPLIT FROM 5.3.3] Verify that untrusted input must not be applied via innerHTML, document.write, or other properties or functions that render HTML. Instead, use createTextNode, textContent, and similar safe functions that do not render HTML and only render content as text.
> 
> What about when the user enters markdown or HTML (which is then filtered using some library). Does "filtered user input" count as "untrusted input"?
> 
> I would consider it is not allowed by this wording requirement but should be allowed. Should we add some provision for when the input has been properly filtered?

This is discussion about sanitization. See my comment here: https://github.com/OWASP/ASVS/issues/2468#issuecomment-2549641895

elarlang:
@randomstuff - I merged the PR, if you have further some concerns or proposals, please write those to related issue and then we can re-open it (if needed)

Issue #2481, opened 2024-12-18
Clarify CSRF requirement 50.4.1


tghosth:
Current text:

| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **50.4.1** | [MODIFIED, MOVED FROM 4.2.2, MERGED FROM 13.2.3] Verify that the application defends against Cross-Site Request Forgery (CSRF) attacks to protect authenticated or sensitive public functionality, using the development framework's built-in anti-CSRF functionality or CSRF tokens, along with additional defense-in-depth measures. | ✓ | ✓ | ✓ | 352 |

We also have:

| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **50.4.3** | [ADDED, SPLIT FROM 14.5.3] Verify that the Origin header field is validated against a defined list of allowed origins to match the desired Cross-Origin Resource Sharing (CORS) policy. | ✓ | ✓ | ✓ | 346 |

Points to change:

* Merge origin validation from 50.4.3 into 50.4.1, as discussed here: https://github.com/OWASP/ASVS/issues/876#issuecomment-2550873677
* Mention triggering a CORS preflight

Potential wording:

_[MODIFIED, MOVED FROM 4.2.2, MERGED FROM 13.2.3] Verify that the application defends against Cross-Site Request Forgery (CSRF) attacks to protect authenticated or sensitive public functionality. This should use the development framework's built-in anti-CSRF functionality or CSRF tokens, along with additional defense-in-depth measures such as validating the origin header. Where cross-origin requests are used, this should always trigger a CORS preflight request first._

jmanico:
> [MODIFIED, MOVED FROM 4.2.2, MERGED FROM 13.2.3] Verify that the application defends against Cross-Site Request Forgery (CSRF) attacks to protect authenticated or sensitive public functionality. This should use the development framework's built-in anti-CSRF functionality or CSRF tokens, along with additional defense-in-depth measures such as validating the origin header. Where cross-origin requests are used, this should always trigger a CORS preflight request first.

Notes:

* Simple cross-origin javscript based requests (XHR, fetch) do not trigger pre-flight.
* Cross origin requests from non-JavaScript sources (like HTML forms) do not trigger pre-flight or SOP.

randomstuff:
[...]

PR #2483, opened 2024-12-18
Remove 13.2.5 to resolve #2472

### LINKED ISSUE #2472: Clarify reasoning behind 13.2.5
> _This this is the problem to fix not to say 13.2.5 is not valid defense against pre-preflight. 13.2.5 is important requirement for CORS, if there is another meaning in another context, I prefer to make a separate requirement for the other context with a clear definition, what other risk it mitigates._

https://github.com/OWASP/ASVS/issues/876#issuecomment-2545170670

> _Well I would actually argue that 13.2.5 is defence in depth for CSRF because the correct fix is what is in 50.4.1. It is irrelevant for CORS because again even if you manage to send the request, you cannot read the response._

https://github.com/OWASP/ASVS/issues/876#issuecomment-2545197336

We should clarify the reason for 13.2.5

| **13.2.5** | Verify that REST services explicitly check the incoming Content-Type to be the expected one, such as application/xml or application/json. | | ✓ | ✓ | 436 |

tghosth:
This Pull Request relates to issue #2472


elarlang:
Feels like endless fight-back and denial :)

> From CORS point of view it gets merged into "CSRF" requirement.

It gets merged into V50.4.1, will be solved with #2481

PR #2488, opened 2024-12-26
Resolve #2476 by rearranging input validation requirements

### LINKED ISSUE #2476: Should Input Validation be Level 1
We currently have an explanation of the thinking around L1 [which can be found here](https://appsecg.host/level1).

There are a few major areas where we have differing opinions about whether a particular type of control falls into the category of L1 for version 5.0.

This issue is to discuss input validation requirements from the current [section 5.1](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x13-V5-Validation-Sanitization-Encoding.md#v51-input-validation)

The currently disputed requirements are 5.1.3, 5.1.4,and 5.1.7 (ignore the level marking below). 5.1.5 is a different discussion and it is agreed that the others in this section are not L1)

| **5.1.3** | [MODIFIED] Verify that all input is validated using positive validation, against an allowed list of values, patterns or ranges to enforce business or functional expectations for that input. | ✓ | ✓ | ✓ | 20 |
| **5.1.4** | [MODIFIED, SPLIT TO 5.1.7] Verify that data items with an expected structure are validated according to the pre-defined rules. | ✓ | ✓ | ✓ | 20 |
| **5.1.7** | [ADDED, SPLIT FROM 5.1.4] Verify that the application ensures that combinations of related data items are reasonable according to the pre-defined rules. | ✓ | ✓ | ✓ | 20 |

I personally believe that these requirements **do not** belong in L1 for the following reasons:

- In all except very specific cases, input validation is a defense in depth issue. There will usually be another first line control (e.g. output encoding, parametrization, context specific sanitization, etc) which is needed to prevent a security vulnerability from being exploited. For instances where a specific mechanism is not mentioned, we already have [5.2.2](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x13-V5-Validation-Sanitization-Encoding.md#v52-sanitization-and-sandboxing) to cover those.
- As a second layer of defense, there is less risk reduction compared to the first line control.
- Comprehensive input validation is very labour intensive, to be done correctly it requires a full mapping of all possible inputs and their business reasoning. This is a problem because:
  - It is hard to justify the effort for a second line control at the lowest level of maturity.
  - In general, it violates the idea that L1 should have a low barrier to entry.
  - Security cannot provide guidance for this alone but will rather need the business/product management team to agree to restrict inputs for which there is likely to be pushback.
  - If I am honest, I am not sure how exactly this would be verified anyway? Certainly it would be problematic to verify "black box" style.

tghosth:
This Pull Request relates to issue #2476

### DIFF
--- a/5.0/en/0x13-V5-Validation-Sanitization-Encoding.md
+++ b/5.0/en/0x13-V5-Validation-Sanitization-Encoding.md
@@ -18,11 +18,10 @@ The "untrusted client" term here refers to client-side technologies that render 
 
 | # | Description | L1 | L2 | L3 | CWE |
 | :---: | :--- | :---: | :---: | :---: | :---: |
-| **1.5.1** | [MODIFIED, SPLIT TO 1.5.5, LEVEL L2 > L1] Verify that input validation rules define how to check the validity of data items against an expected structure. This could be common data formats such as credit card numbers, e-mail addresses, telephone numbers, or it could be an internal data format. | ✓ | ✓ | ✓ | 20 |
+| **1.5.1** | [SPLIT TO 1.11.5, 1.11.6] | | | | |
 | **1.5.2** | [DELETED, MERGED TO 5.5.3] | | | | |
 | **1.5.3** | [MOVED TO 5.6.2] | | | | |
 | **1.5.4** | [MOVED TO 5.6.3] | | | | |
-| **1.5.5** | [ADDED, SPLIT FROM 1.5.1] Verify that input validation rules are documented and define how to ensure the logical and contextual consistency of combined data items, such as checking that suburb and zip code match. | ✓ | ✓ | ✓ | 20 |
 
 ## V5.1 Input Validation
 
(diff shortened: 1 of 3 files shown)


elarlang:
Attempt 416 to get it done :)

Tags - all `MODIFIED, SPLIT FROM` should be `ADDED, SPLIT FROM`.

11.3.1 and 11.3.2
*  I don't think "will be made" is correct form, "are made"?
* For L2 - "should" vs "must"?

In general - for this kind of change it is better for trackability to make one move (commit) at a time.

elarlang:
This one also?
> V5.6.2 Verify that the application is designed to enforce input validation at a trusted service layer. While client-side validation improves usability, it must not be relied upon as a security control.

tghosth:
@elarlang I updated the PR.

Once we have finalised the content, I may consider doing a different PR with the changes better organized as separate commits.

elarlang (approved):

tghosth:
Replaced by https://github.com/OWASP/ASVS/pull/2490

PR #2507, opened 2025-01-06
Clarify 2.9.1 to resolve #2463

### LINKED ISSUE #2463: V2.9.1 Cryptographic authentication mechanism, protection against disclosure is not relevant
Current 2.9.1 (talking about smart cards, FIDO devices):

> **2.9.1** [MODIFIED, LEVEL L2 > L3] Verify that the **authentication server** stores the cryptographic keys used in verification are securely and protected against** disclosure**, such as using a Trusted Platform Module (TPM) or Hardware Security Module (HSM), or an OS service that can use this secure storage. 

On the authentication server, these credentials are actually *public keys*. It is therefore not that important to protect them against disclosure. You would not need to store a public key in a TPM or HSM for this usage.

For smart card, these are actually X.509 certificates and protecting them agaisnt disclosure often does not make sense at all.

If I understand this requirement correctly, I think it is not needed.

tghosth:
This Pull Request relates to issue #2463

### DIFF
--- a/5.0/en/0x11-V2-Authentication.md
+++ b/5.0/en/0x11-V2-Authentication.md
@@ -197,7 +197,7 @@ The requirements in this section mostly relate to section [5.1.7.2](https://page
 
 | # | Description | L1 | L2 | L3 | CWE |
 | :---: | :--- | :---: | :---: | :---: | :---: |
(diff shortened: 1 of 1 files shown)


randomstuff:
LGTM.

tghosth:
Ok so I think this is ready to merge

Issue #2498, opened 2025-01-02
Cryptography - suggested modification of 6.5.5


randomstuff:
Currently we have:

> 6.5.5 [ADDED] Verify that any authenticated signatures are operating in encrypt-then-MAC or encrypt-then-hash modes as required.

Proposed by Bart Preneel:

> 6.5.5 [ADDED] Verify that any combination of an encryption algorithm and a MAC algorithm is operating in encrypt-then-MAC modes as required.

Comment:

> Digital signatures are rarely combined with encryption. I would not use the term signatures for a MAC algorithm.

randomstuff:
This would change the requirement completely.

Is there any need for the original requirement? Is this used in practice?

My problem with the proposed requirement is that TLS 1.2 uses MAC-then-encrypt for non AEAD (eg.  TLS_AES_128_CCM_SHA256) (unless when using some [TLS extension](https://datatracker.ietf.org/doc/html/rfc7366) with does not appear to be used in practice). If we have this requirement, should some provision be made for legacy/compatibility (in particular with TLS 1.2)?

jmanico:
Again, signatures being used incorrectly again. We really want to get our terminology right. I suggest we fix this for 5.0.

randomstuff:
I just noticed the "authenticated signatures" wording, which is weird. I think we are talking both about authenticated encryption and signcryption here (?).

unprovable:
"authenticated signatures" is a valid phrase, so I'm not sure what the issue is in this regard? But you're right they shouldn't be conflated with MACs, which they aren't. I think this part is supposed to be talking about AEAD and making sure this is present in symmetric cipher use.

jmanico:
Thanks for this clarification, I’m good. Makes sense to me.

tghosth:
What do you think the next stage here is @randomstuff ?

randomstuff:
I think we can use the proposed wording:

> 6.5.5 [ADDED] Verify that any combination of an encryption algorithm and a MAC algorithm is operating in encrypt-then-MAC modes as required.

MAC-then-encrypt might be required for compatibility with (very) old TLS implementations (eg. CBC ciphersuites) but I don't think we really need to handle this. If required, we could add:

> As an exception, MAC-then-encrypt might be used in TLS when communicating with legacy clients.

tghosth:
Ok so please can you open a PR

tghosth:
Closed by #2533

PR #2533, opened 2025-01-16
Tweak wording for encrypt-then-MAC (6.5.5)

randomstuff:
Relates to #2498

I removed "as required" because I don't think this bring anything.


jmanico (commented):
there is some extra text here, delete "[ADDED] 6.5.5 "

tghosth (approved):

PR #2534, opened 2025-01-16
Disapprove usage of CCM-8 to resolve #2485

### LINKED ISSUE #2485: 6.2.9 and CCM-8
Since https://github.com/OWASP/ASVS/pull/2482, 6.2.9 is:

> **6.2.9** [ADDED] Verify that all cryptographic primitives utilize a minimum of 128-bits of security based on the algorithm, key size, and configuration. For example, a 256-bit ECC key provides roughly 128 bits of security where RSA requires a 3072-bit key to achieve 128 bits of security.

CCM-8 is still listed in the approved algorithms. However, its authentication tag only has 64 bits of security and does not respect this minimum of 128-bits of security.

There seems to be a conflict between these. How should be fix this?

Possible options:

* add some exception in 6.2.9 in order to allow CCM-8 in certain contexts ("unless some additional measure to prevent forgery of messages is used");
* do not approve CCM-8.

I don't have a strong opinion on this and I don't know if CCM-8 is really used/useful nowadays. The safest choice seems to disapprove CCM-8. Unless/until someone come forwards with further input, I would suggest taking this approach.

See as well discussion in https://github.com/OWASP/ASVS/issues/2413 for more info about CCM-8.

randomstuff:
Relates to #2485


tghosth:
See my comment, https://github.com/OWASP/ASVS/issues/2485#issuecomment-2600866737

tghosth (approved):

PR #2399, opened 2024-11-21
Crypto appendix: allow HMAC-SHA-1

randomstuff:

PR #2536, opened 2025-01-19
Remove duplicates and tidy tags to resolve #2523

### LINKED ISSUE #2523: Redundant requirement about secure hash functions
6.6.1 and 6.6.3 look very much the same to me:

| **6.6.1** | [ADDED] Verify that <mark>only approved hash functions</mark> are used for general cryptographic use cases, including digital signatures, HMAC, KDF, and random bit generation. | | ✓ | ✓ | |
| **6.6.3** | [ADDED, SPLIT FROM 6.2.5] Verify that cryptographic systems <mark>avoid the use of disallowed hash functions</mark>, such as MD5, SHA-1, or any other insecure hash functions, for any cryptographic purpose. | ✓ | ✓ | ✓ | |

A disapproved hash function cannot be approved right? So if we check that they are approved, we also check they are not disallowed?

By the way should be say "cryptographic hash functions" everywhere. For example, SipHash might very well be an approved hash function but is not an approved *cryptographic* hash functions. (Sure, we are in the cryptography chapter but still …).

tghosth:
This Pull Request relates to issue #2523

### DIFF
--- a/5.0/en/0x14-V6-Cryptography.md
+++ b/5.0/en/0x14-V6-Cryptography.md
@@ -51,9 +51,9 @@ Although this section is not easily penetration tested, developers should consid
 | :---: | :--- | :---: | :---: | :---: | :---: |
 | **6.2.1** | [MODIFIED] Verify that all cryptographic modules fail securely, and errors are handled in a way that does not enable vulnerabilities, such as Padding Oracle attacks. | ✓ | ✓ | ✓ | 310 |
 | **6.2.2** | Verify that industry proven or government approved cryptographic algorithms, modes, and libraries are used, instead of custom coded cryptography. | | ✓ | ✓ | 327 |
-| **6.2.3** | [DELETED, COVERED BY 6.5.1, 6.5.2, 6.6.3] | | | | |
+| **6.2.3** | [DELETED, COVERED BY 6.5.1, 6.5.2, 6.6.1] | | | | |
 | **6.2.4** | [MODIFIED, MERGED FROM 1.6.3] Verify that the application is designed with crypto agility such that random number, authenticated encryption, MAC,  or hashing algorithms, key lengths, rounds, ciphers or modes can be reconfigured, upgraded, or swapped at any time, to protect against cryptographic breaks. Similarly, it must also be possible to replace keys and passwords and re-encrypt data. This should allow for seamless upgrades to post-quantum cryptography (PQC), once high-assurance implementations of approved PQC schemes or standards are widely available. | | ✓ | ✓ | 320 |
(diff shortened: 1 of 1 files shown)


tghosth:
Fixed those 3 issues @elarlang

5.0/en/0x14-V6-Cryptography.md
```diff
@@ -96,10 +96,9 @@ Cryptographic hashes are used in a wide variety of cryptographic protocols, such
 
 | # | Description | L1 | L2 | L3 | CWE |
 | :---: | :--- | :---: | :---: | :---: | :---: |
-| **6.6.1** | [ADDED] Verify that only approved hash functions are used for general cryptographic use cases, including digital signatures, HMAC, KDF, and random bit generation. | | ✓ | ✓ | |
+| **6.6.1** | [ADDED, SPLIT FROM 6.2.5, COVERS 6.2.3] Verify that only approved hash functions are used for general cryptographic use cases, including digital signatures, HMAC, KDF, and random bit generation. Disallowed hash functions, such as MD5, SHA-1, must not be used for any cryptographic purpose. | | ✓ | ✓ | |
```
randomstuff, line 99: > Disallowed hash functions, such as MD5, SHA-1, must not be used for any cryptographic purpose.

There is some inconsitency here as HMAC-SHA-1 is currently allowed.

randomstuff: As per #2399

randomstuff: Therefore I would suggest using:

> [ADDED, SPLIT FROM 6.2.5, COVERS 6.2.3] Verify that only approved hash functions are used for general cryptographic use cases, including digital signatures, HMAC, KDF, and random bit generation. Disallowed hash functions, such as MD5, must not be used for any cryptographic purpose.

tghosth: Is there any situation where MD5 would be allowed? Can you open a PR @randomstuff ?

jmanico:
While sha1 has collisions HMAC-sha1 does not. HMAC-sha1 is only a security problem as quantum computers come online.

randomstuff:
@jmanico, Yes my point was that we are saying in one place that « SHA-1 must not be used for **any** cryptographic purpose » but in some other place that using SHA-1 as part of the HMAC construct is allowed.

5.0/en/0x14-V6-Cryptography.md
randomstuff: @tghosth, this is #2546

Issue #2492, opened 2025-01-02
13.6.2 move into V50 (if not merged into V50.4.1)


elarlang:
Current requirement 13.6.2:
| # | Description | L1 | L2 | L3 | CWE |
| :---: | :--- | :---: | :---: | :---: | :---: |
| **13.6.2** | [MODIFIED, MOVED FROM 13.2.1] Verify that HTTP requests using the HEAD, OPTIONS, TRACE or GET verb do not modify any backend data structure or perform any state-changing actions. These requests are safe methods and should therefore not have any side effects. | ✓ | ✓ | ✓ | 650 |

I think this is browser-based attack scenario specific and a pre-condition to have defense against (so-callsd) CSRF attacks in place.

The option is to merge into to the "csrf" requirement or move to the same section.

Related comment and issue: https://github.com/OWASP/ASVS/issues/2481#issuecomment-2567821725

tghosth:
So the question is, is CSRF the only security risk which this requirement addresses?

In principle, there is a general assumption that HEAD, OPTIONS, TRACE or GET verb do not modify any backend data structure or perform any state-changing actions. However, I am not sure there is a specific security risk outside of CSRF.

This requirement seems to have been added as a result of the discussion in #672 which included people such as @Sjord, @VincentDS, and @Someniak.

Based on that discussion I think the only risk is CSRF and therefore I propose that this should be merged into the CSRF requirement (currently being worked on in #2481).

I am open to other points of view though.

tghosth:
This will get moved into v50.4.4 as part of #2481

PR #2548, opened 2025-01-26
2481-clarify-csrf-requirement-5041

tghosth:
This Pull Request relates to issues: 

* #2481 
* #2472
* #2492


elarlang:
I would like to see tag-connection from v4.0.3-13.2.3 to v5.0.be-50.4.3.

I think we can use `COVERED BY 50.4.1, 50.4.3` for 13.2.3 and `COVERS 13.2.3` for 50.4.1 and 50.4.3.

tghosth:
Ok so to be clear the reason for this would be that 4.2.2 and 13.2.3 were duplicates of each other?

elarlang:
Yes, in practice they were. But if we take into context, that 13.2.3 was in API section and provides validating the `Origin` request header field as an option, the intent there in my opinion is also to cover the restful API area. It is not just well written, but I would like to make the connection, that the 13.2.3 now is covered by 50.4.3 as well.

tghosth:
Ok cool, do you want to keep them as separate commits or merge them into a single commit?

elarlang:
if we do requirement content changes, it's always better to have separate commits

elarlang (approved):

Issue #2398, opened 2024-11-21
Appendix Crypto - Allowed mechanisms and requirement levels


randomstuff:
Currently the crypto appendix as table as follow:

| Symmetric Key Algorithms | Reference | L1 | L2 | L3 |
|--|--|--|--|--|
| AES-256 | [FIPS 197](https://csrc.nist.gov/pubs/fips/197/final) | | ✓ | ✓ |
| ChaCha20 | [RFC 8439](https://www.rfc-editor.org/info/rfc8439) | | ✓ | ✓ |
| AES-192 | [FIPS 197](https://csrc.nist.gov/pubs/fips/197/final) | | ✓ | ✓ |
| AES-128 | [FIPS 197](https://csrc.nist.gov/pubs/fips/197/final) | ✓ | ✓ | ✓ |

The relation between the check marks and the levels does not make much sense. Why would AES-256 be OK for L2 and L3 but nor for L1? To make sense, it would have to be the other way around:

| Symmetric Key Algorithms | Reference | L1 | L2 | L3 |
|--|--|--|--|--|
| AES-256 | [FIPS 197](https://csrc.nist.gov/pubs/fips/197/final) | ✓ | ✓ | ✓ |
| ChaCha20 | [RFC 8439](https://www.rfc-editor.org/info/rfc8439) | ✓ | ✓ | ✓ |
| AES-192 | [FIPS 197](https://csrc.nist.gov/pubs/fips/197/final) | ✓ | ✓ | ✓ |
| AES-128 | [FIPS 197](https://csrc.nist.gov/pubs/fips/197/final) | ✓ | | |

All these should be adjusted, I think.

danielcuthbert:
In all honesty, I don't see us needing any levels here. It should apply to all no?

randomstuff:
> It should apply to all no?

Yes possibly. Removing the levels here might be an option.

Would it make sense to have stricter requirements in this regard for L1 applications than for L3/financial/healthcare ones? In the sense, that if you already use some somewhat sub-par cipher/MAC/signature for L1 applications, it might be OK for now but you should really use something better for L3.

For example, the appendix currently says: that EdDSA (Ed25519, Ed448) is allowed. On the other hand, FAPI 2.0 says "EdDSA (using the Ed25519 variant)". Should the appendix align with FAPI 2.0 in this regard? Only for L3?

So there are two options:

1. remove level indications in these list;
2. fix them.

randomstuff:
[...]

PR #2573, opened 2025-02-05
Crypto: define mechansism status

randomstuff:
This PR is not intended to be merged for now. It is *proof-of-concept* for allowed/legacy/disallowed status of cryptographic mechanisms (#2398) and is intended to serve as a reference for discussing this option and a possible presentation style. The exact content would need to be verified/adjusted.

:warning: Some important wording from the original content might be missing and might need to be brought back.

### DIFF
--- a/5.0/en/0x97-Appendix-V_Cryptography.md
+++ b/5.0/en/0x97-Appendix-V_Cryptography.md
@@ -2,13 +2,24 @@
 
 The "Cryptography" chapter goes beyond simply defining best practices. It aims to enhance understanding of cryptography principles and encourage the adoption of more resilient, modern security methods. This appendix provides detailed technical information regarding each requirement, complementing the overarching standards outlined in the "Cryptography" chapter.
 
+This appendix defines the level of approval for different cryptographic mechanisms:
+
+* **Approved** (A) mechanisms can be used in applications.
+* **Legacy** mechanisms (L) should not be used in applications but might still be used for compatibility with existing legacy applications or code onyly. While the usage of such these mechanisms is currently not considered to be a vulnerability in itself, they should be replaced by more secure and future-proof mechanisms as soon as possible.
+* **Disallowed** mechanisms (D) must not be used because they are currently considered broken or do not provide sufficient security.
+
+This list may be overloaded in the context of a given applications for different reasons including:
+
+* new evolutions in the field of cryptography;
+* compliance with some regulation.
+
 ## Algorithms (V11.2)
 
-### Equivalent Strengths of Cryptographic Parameters
+## Equivalent Strengths of Cryptographic Parameters
 
 The relative security strengths for various cryptographic systems are in this table (from [NIST SP 800-57 Part 1](https://csrc.nist.gov/pubs/sp/800/57/pt1/r5/final), p.71):
 
-| Security Strength | Symmetric Key Algorithms | Finite Field Cryptography (DSA, DH, MQV) | Integer Factorisation Cryptography (RSA) | Elliptic Curve Cryptography (ECDSA, EdDSA, DH, MQV) |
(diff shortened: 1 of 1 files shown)


randomstuff:
I am highlighting here some changes I made along the way:

* mention PKDF2-SHA1 (legacy?)
* removed KMAC from the list of hash function because it is a MAC
* "Diffie-Hellman" → "Finite Field Diffie-Hellman"

tghosth:
@danielcuthbert @unprovable do you have any thoughts on the suggestions from @randomstuff ?

unprovable:
Sure, some thoughts:
* Yes it is required to allow MD5 and SHA1 in many old and regulated environments. If you don't permit this, then few financial institutions can ever be ASVS certified. It's short sighted to remove a reasonable control.
* You added in PBKDF2-SHA1 which is a legacy control for verifying old hashes. If we can permit this, then we can permit the text saying that old hashes are verification only (notify that SHA1 is treated with similar concern as we do MD5 etc.)
* For any PBKDF you must state the number of iteration that are considered safe (not sure if this is tue case already here, but it is required for all the other standards we maintain).
* The other two points are fine as I far as I can see.

Hope this helps! M.

danielcuthbert:
I'm with @unprovable here. The use of MD5 and SHA1 for verification purposes only and I really stress this, VERIFICATION ONLY, is required for many orgs that still use this. I can think of many in ICS/CNI/Financial who do use MD5 in this way and want to be Level 2 and above.

jmanico:
> * Yes it is required to allow MD5 and SHA1 for verification in many old and regulated environments. If you don't permit this, then few financial institutions can ever be ASVS certified. It's short sighted to remove a reasonable control.

I think it's good that they can't be ASVS certified. They should stop using md5. I don't think, with respect my friends, that this is a good reason to support out of date crypto.

> * You added in PBKDF2-SHA1 which is a legacy control for verifying old hashes. If we can permit this, then we can permit the text saying that old hashes are verification only (notify that SHA1 is treated with similar concern as we do MD5 etc.)

For the legacy section only, **and not in core requirements**, I think is ok.

> * For any PBKDF you must state the number of iteration that are considered safe (not sure if this is tue case already here, but it is required for all the other standards we maintain).

Agreed. Depending on what it's used for (key generation vs password storage) these numbers are usually different per use.

randomstuff:
> I think it's good that they can't be ASVS certified. They should stop using md5. I don't think, with respect my friends, that this is a good reason to support out of date crypto.

Yes, that was basically my thinking as well.

As an counter-example, maybe `FooAcme` wants to be ASVS Level 3 certifed but don't want to hash their passwords. That's their problem to solve, it does not mean that the requirement about password hashing should go away.

> old hashes are verification only

Isn't “we don't generate new MD5s (or other broken hash function) hashes but still accept them for verification” dubious/dangerous? If you accept old broken hashes, it won't stop an attacker from generating new ones. Isn't your verification is mostly useless?

(In the general case. It probably depends based on the way hash is used i.e. it's probably not as bad for password hashing than for MAC or signature verification?)

randomstuff:
@unprovable 

> You added in PBKDF2-SHA1 which is a legacy control for verifying old hashes. If we can permit this, then we can permit the text saying that old hashes are verification only (notify that SHA1 is treated with similar concern as we do MD5 etc.)

Do we? I had the understanding that HMAC-SHA-1 (and therefore PBKDF2-HMAC-SHA-1) was still considered not-broken in contrast to plain SHA-1 (eg. as used in digital signatures).

tghosth:
Hi @randomstuff, please can you fix the conflict on this PR and then I will ask daniel and mark to take another look

randomstuff:
Rebased on top of master. Ping @danielcuthbert and @unprovable for feedback.

tghosth (approved):

Issue #2590, opened 2025-02-09
Is 2.9.3 a duplicate of 6.5.2 (crypto algorithms)


tghosth:
**Chapter text for 2.9:**

Cryptographic authentication mechanism include smart cards or FIDO keys, where the user has to plug in or pair the cryptographic device to the computer to complete authentication. The authenticatoin server will send a challenge nonce to the cryptographic device or software, and the device or software calculates a response based upon a securely stored cryptographic key.

The requirements for single-factor cryptographic devices and software, and multi-factor cryptographic devices and software are the same, as verification of the cryptographic device proves possession of the authentication factor.

**Relevant requirements for this issue:**

| # | Description | Level | CWE |
| :---: | :--- | :---: | :---: |
| **2.9.3** | [MODIFIED, LEVEL L2 > L3] Verify that approved cryptographic algorithms are used in the generation, seeding, and verification of the cryptographic keys. | 3 | 327 |
| **6.5.2** | [ADDED, SPLIT FROM 6.2.5, COVERS 6.2.3, LEVEL L2 > L1] Verify that only secure, authenticated ciphers and modes such as AES with GCM are used. | 1 | 326 |

Should 2.9.3 be merged into 6.5.2? Feels like it is covered there. Is this something that absolutely needs its own separate requirement.

@elarlang @jmanico  @randomstuff

randomstuff:
6.5.2 is actually about *encryption* (cipher). Smart card and (I think) FIDO keys use digital signatures so they are not really covered by 6.5.2. I agree that 2.9.3 *should* be covered by some requirement in 6.x however. We don't really have requirements about digital signatures in V6 however.

jmanico:
I think these can absolutely be merged. There is likely a lot of other merging possible in the crypto section. 

Perhaps for this one we may want to go this route:


Verify that only approved cryptographic algorithms and modes of operation are used for key generation, seeding, verification, and encryption.

randomstuff:
> Verify that only approved cryptographic algorithms and modes of operation are used for key generation, seeding, verification, and encryption.

Yes, I agree that such a change makes sense (it we want to have a more exhaustive coverage of cryptographic topics without delving too deep into the details).

I think we would have to add "digital signature" and "MAC" as well (there is "verification" in your proposition but not generation of these). It could make sense sense to include Key Exchange as well (which is currently in 6.7.1).

jmanico:
So how about:

[...]

PR #2624, opened 2025-02-19
Focus 6.2.2 and split the algorithms part to other reqs as part of #2590

tghosth:
This Pull Request relates to issue #2590 but does not fully resolve it.

A 2nd PR should be opened to handle the original problem in #2590 which was that 2.9.3 was a duplicate.


jmanico (approved):
Great working bring this home everyone

Issue #2621, opened 2025-02-18
OAuth 2.0 V51 - Sanity Check Comments/Suggestions for v.5.0


csfreak92:
**Chapter V51 - OAuth 2.0** 

Here are a few things I observed reviewing this chapter for v.5.0 for sanity check related to #2582: 

_V51.7 Consent Management_
Needs some paragraph describing this section. For consistency across all subsections in ASVS.

I also saw some requirements that need some grammatical changes which I will reference in this issue in a PR later.

csfreak92:
For this intro paragraph, I am not sure what must be written here. I would lean for the folks who are more expert in this set of requirements for their PR.

tghosth:
@randomstuff @TobiasAhnoff @elarlang could you go through the chapter and section text here to make sure it is ok?

elarlang:
The last one (V51.7) requires some section text, but otherwise, all 3 of us have worked with the text, and from my point of view, it makes sense to have external eyes on that.

If there is no further action to be expected from this issue, it can be closed. Parallel reviews are in place anyway.

randomstuff:
Maybe something like:

> These requirement cover the verification of the user's consent by the authorization server. Without proper user consent verification, a malicious actor could easily obtain permissions on the user behalf through spoofing or social-engineering.

PR #2625, opened 2025-02-19
Update 0x51-V51-OAuth2.md

csfreak92:
Some minor grammatical changes that I spotted.

This Pull Request relates to issue #2621 just for the small grammatical changes that I spotted. I did not yet add anything for the V51.7 Consent Management section description as I am not sure what it really does. I will leave it to someone more capable and more of an expert on it.


elarlang:
Can you please explain reasoning, rationale and link to grammar or stylebook, why we should use "Pushed Authorization Requests (PAR)" instead of "pushed authorization requests (PAR)"?

Related to issue #1875

Yedfree (approved):

csfreak92:
The reasoning is, having consistency with some terms like why would we call OAuth 2.0 instead of oauth 2.0, same as with Demonstration of Proof of Possession (DPoP) where it can be demonstration of proof of possession? It seems not consistent to camel-case some items that have acronyms next to them then we don't do them to others. It makes it reading a bit harder. 

Then there are also had some acronyms like RP but we never mentioned what it is. It could be in the glossary and intro text but not a lot of people read that. To be concise, whenever an acronym or abbreviation like RP is there or RO, they have to be spelled out with relying party and resource owner, respectively.

elarlang:
OAuth is a clear name and does not belong to this discussion, DPoP is from the same league.

Consistency is a goal and issue to solve and we have a separate issue for that. If we make those changes, I prefer to have clear reasons and facts to do so, otherwise,e we keep changing them back and forth without going anywhere. So we need to have a result from #1875 first and that is why I asked, if is there any style guide or grammar behind those changes.

jmanico:
Pushed Authorization Requests (PAR) is the proper english since this is a proper noun, a specific standard, or a formalized concept in a formal document.

elarlang:
I made it lowercase, because the RFC that describes PAR uses it that way https://datatracker.ietf.org/doc/html/rfc9126.

I'm not the person who has too much to say about grammar, my question is - what is the chance that the RFC gets it wrong? And it is wrong and we are going to use another version, it must be clear and documented.

tghosth:
I emailed the RFC authors so we'll see if we get anything 🙃

In the meantime, it seems like the RFC is the best source of truth we have.

At some point we will have to resolve #1875 during which we may revisit this but for now I would suggest we go with the RFC.

elarlang:
I would like to have a "clear stage" for V51 for next 10 days because I need to do changes and going to re-order requirements. So if there is no quick solution for this PR, I'll close it out.

tghosth (approved):

tghosth:
Ok, I am going to merge this in with PAR aligned to the RFC capitalization. If there are concerns let's open a separate issue. Thanks again @csfreak92 :)

PR #2626, opened 2025-02-19
Clarify input val to resolve #2614

### LINKED ISSUE #2614: Move 11.1.3 to 11.3.5
Do you think this might belong in the V11.3 Input Validation instead of the business logic section? Perhaps 11.3.5?

 _11.1.3 [MODIFIED, MERGED FROM 11.1.5] Verify that business logic limits and validations are implemented as per the application's documentation, to avoid business logic flaws being exploited such as buying items for a negative amount._

tghosth:
This Pull Request relates to issue #2614

Would appreciate approvals from both @elarlang and @jmanico on the changes here.


jmanico (approved):
I still suggest making this "all input" and drop the global making this l1. I think input validation should be universal for all apps, but am totally ok if you disagree, hence the approval. :)

elarlang (approved):

PR #2639, opened 2025-02-20
Resolve #2628 by better focusing 7.4.1

### LINKED ISSUE #2628: 7.4.1 split
| # | Description | Level | CWE |
| **7.4.1** | Verify that a generic message is shown when an unexpected or security sensitive error occurs, potentially with a unique ID which support personnel can use to investigate. | 1 | 210 |

Here are 2 requirements:
* do not show a technical error message to output
* show unique ID for support

First, this "potentially" is... noise.

I think the first part is worth being a level 1 and this unique ID can be level 3.

Just showing a unique ID does not make sense, it requires support from the company/organization process, and it is a separate topic from "do not show details to output", so I propose to split this.

tghosth:
This Pull Request relates to issue #2628

### DIFF
--- a/5.0/en/0x15-V7-Error-Logging.md
+++ b/5.0/en/0x15-V7-Error-Logging.md
@@ -68,7 +68,7 @@ The purpose of error handling is to ensure the application fails gracefully and 
 
 | # | Description | Level | CWE |
 | :---: | :--- | :---: | :---: |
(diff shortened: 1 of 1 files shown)


elarlang (commented):
tag `MODIFIED`, as it is not just a grammar change, otherwise ok

tghosth:
fair one @elarlang I fixed that

elarlang (approved):

Issue #2109, opened 2024-09-23
V51 - OAuth - confidential client


elarlang:
Spin-off from https://github.com/OWASP/ASVS/issues/2040#issuecomment-2338933353

> Verify that confidential clients must use client authentication in refresh token requests. (L1, L2, L3)

Isn't it the point for confidential client to use client authentication?

If we require it from Level 1 from confidential client but allow public clients without any restrictions, I would say there is a conflict.

Is it required only for refresh token request or to any request to token endpoint?

randomstuff:
I believe the idea here is that that some authorization server implementations could assume that presenting the refresh token is enough for refreshing the token (as presenting the access token is enough for using the service). This is requirement is indeed valid for all requests to the token endpoint.

A more precise wording of this requirement could be something like: ”Verify that, for using refresh tokens emitted to a confidential client, the authorization server requires authentication *of this client*.“

elarlang:
> Isn't it the point for confidential client to use client authentication?

My point is - can you use confidential client without authentication for the token endpoint?

> ... for using refresh tokens emitted

but asking the refresh token with token request?

TobiasAhnoff:
> My point is - can you use confidential client without authentication for the token endpoint?

It should not be possible, but not all AS implementations follows the RFC which states that confidential clients must be authenticated, both when using refresh-tokens, authorization codes or client-credentials.
See in example https://www.rfc-editor.org/rfc/rfc6749.html#section-4.1.3

This is important for a secure OAuth implementation, but perhaps not a widespread issue, except if using a custom built AS or flow (not "well-known industry standard"). Still, I think it is a good level 1 requirement for ASVS.

It could also me written to not just focus on refresh-tokens, like this

> Verify that all confidential clients are authenticated for all token requests. (L1, L2, L3) 

There is no conflict with public clients, since the requirement only applies to confidential clients.

elarlang:
Ok, the definition says "capable", not doing so https://datatracker.ietf.org/doc/html/rfc6749#section-2.1
>    confidential - Clients capable of maintaining the confidentiality of their credentials (e.g., client implemented on a secure server with restricted access to the client credentials), or capable of secure client authentication using other means.

So from that perspective it makes sense.

Should we add "authorization server" in the requirement?
> Verify that all confidential clients are authenticated for all token requests **to the authorization server**.

elarlang:
At the moment it goes to the document as:
| # | Description | L1 | L2 | L3 |
[...]

PR #2648, opened 2025-02-24
agreed levels between Josh and Elar, for review

elarlang:
Level rework.

### DIFF
--- a/5.0/en/0x51-V51-OAuth2.md
+++ b/5.0/en/0x51-V51-OAuth2.md
@@ -48,8 +48,8 @@ These requirements cover generic architectural requirements that apply to all ap
 
 | # | Description | Level |
 | :---: | :--- | :---: |
-| **51.1.1** | [ADDED] Verify that tokens are only sent to components that strictly need them. For example, when using a backend-for-frontend pattern for browser-based JavaScript applications, access and refresh tokens shall only be accessible for the backend. | 1 |
-| **51.1.2** | [ADDED] Verify that the client only accepts values from the authorization server (such as the authorization code or ID token) if these values result from an authorization flow that was initiated by the same user agent session and transaction. This requires that client-generated secrets, such as the proof key for code exchange (PKCE) 'code_verifier', 'state' or OIDC 'nonce' are not guessable, are specific to the transaction, and are securely bound to both the client and the user agent session in which the transaction was started. | 1 |
+| **51.1.1** | [ADDED] Verify that tokens are only sent to components that strictly need them. For example, when using a backend-for-frontend pattern for browser-based JavaScript applications, access and refresh tokens shall only be accessible for the backend. | 2 |
+| **51.1.2** | [ADDED] Verify that the client only accepts values from the authorization server (such as the authorization code or ID token) if these values result from an authorization flow that was initiated by the same user agent session and transaction. This requires that client-generated secrets, such as the proof key for code exchange (PKCE) 'code_verifier', 'state' or OIDC 'nonce' are not guessable, are specific to the transaction, and are securely bound to both the client and the user agent session in which the transaction was started. | 2 |
 
 ## V51.2 OAuth Client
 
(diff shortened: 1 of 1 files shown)


tghosth:
@elarlang who is supposed to be reviewing this?

elarlang:
Main focus is to get up-to-date information for OWS, but at the end everyone should review.

tghosth:
Presumably @TobiasAhnoff and @randomstuff are particularly important here, no?

5.0/en/0x51-V51-OAuth2.md
```diff
@@ -48,8 +48,8 @@ These requirements cover generic architectural requirements that apply to all ap
...
+| **51.1.1** | [ADDED] Verify that tokens are only sent to components that strictly need them. For example, when using a backend-for-frontend pattern for browser-based JavaScript applications, access and refresh tokens shall only be accessible for the backend. | 2 |
+| **51.1.2** | [ADDED] Verify that the client only accepts values from the authorization server (such as the authorization code or ID token) if these values result from an authorization flow that was initiated by the same user agent session and transaction. This requires that client-generated secrets, such as the proof key for code exchange (PKCE) 'code_verifier', 'state' or OIDC 'nonce' are not guessable, are specific to the transaction, and are securely bound to both the client and the user agent session in which the transaction was started. | 2 |
```
randomstuff, line 52: In the context of OIDC this prevents OIDC-based login-CSRF through the `state` (or PKCE). I think, I would want it to be L1. I would include CSRF protections (including login-CSRF) in L1.

5.0/en/0x51-V51-OAuth2.md
```diff
@@ -88,31 +88,31 @@ These requirements detail the responsibilities for OAuth authorization servers,
...
 | **51.4.5** | [ADDED] Verify that for a given client, the authorization server only allows the usage of grants that this client needs to use. Note that the grants 'token' (Implicit flow) and 'password' (Resource Owner Password Credentials flow) must no longer be used. | 2 |
 | **51.4.6** | [ADDED] Verify that the authorization server validates redirect URIs based on a client-specific allowlist of pre-registered URIs using exact string comparison. | 1 |
-| **51.4.7** | [ADDED] Verify that confidential client is authenticated for client-to-authorized server backchannel requests such as token requests, pushed authorization requests (PAR), token revocation requests, and token introspection requests. | 1 |
-| **51.4.8** | [ADDED] Verify that the authorization server configuration only assigns the required scopes to the OAuth Client. | 1 |
+| **51.4.7** | [ADDED] Verify that confidential client is authenticated for client-to-authorized server backchannel requests such as token requests, pushed authorization requests (PAR), token revocation requests, and token introspection requests. | 2 |
```
randomstuff, line 95: Making this one L2 is quite weird. If we don't authenticate the confidential client for token request, there is not much much reason to make the client confidential in the first place (the client is not acting as a confidential client) :smile:

randomstuff: Out topic: I just noticed that we include "token introspection requests" here. This endpoint is intended to be used by the RS not the client so I am wondering it we should include this is this requirement. (I'll check if we have discussed this point before).

randomstuff: Note I am the one who suggested including it in the first place https://github.com/OWASP/ASVS/issues/2109#issuecomment-2377422117 :smile:

5.0/en/0x51-V51-OAuth2.md
```diff
@@ -88,31 +88,31 @@ These requirements detail the responsibilities for OAuth authorization servers,
...
 | **51.4.15** | [ADDED] Verify that, for a server-side client (which is not executed on the end-user device), the authorization server ensures that the 'authorization_details' parameter value is from the client backend and that the user has not tampered with it. For example, by requiring the usage of pushed authorization request (PAR) or JWT-secured Authorization Request (JAR). | 3 |
-| **51.4.16** | [ADDED] Verify that if the authorization server supports unauthenticated dynamic client registration, it mitigates the risk of malicious client applications. It must validate client metadata such as any registered URIs, ensure the user's consent and warn the user before processing an authorization request with an untrusted client application. | 2 |
+| **51.4.16** | [ADDED] Verify that if the authorization server supports unauthenticated dynamic client registration, it mitigates the risk of malicious client applications. It must validate client metadata such as any registered URIs, ensure the user's consent and warn the user before processing an authorization request with an untrusted client application. | 3 |
```
randomstuff, line 104: I would keep this one to L2. In this scenario, I think the the risk of abuse by a malicious application is high. I would keep it in L2.

5.0/en/0x51-V51-OAuth2.md
```diff
@@ -88,31 +88,31 @@ These requirements detail the responsibilities for OAuth authorization servers,
...
+| **51.5.1** | [ADDED] Verify that the Client (as the relying party) mitigates ID token replay attacks. For example, by ensuring that the 'nonce' claim in the ID token matches the 'nonce' value sent in the authentication request to the OpenID provider (in OAuth2 refereed to as the authorization request sent to the authorization server). | 2 |
+| **51.5.2** | [ADDED] Verify that the Client uniquely identifies the user from ID token claims, usually the 'sub' claim, which cannot be reassigned to other users (for the scope of an identity provider). | 2 |
+| **51.5.3** | [ADDED] Verify that the client rejects attempts by a malicious authorization server to impersonate another authorization server through authorization server metadata. The client must reject authorization server metadata if the issuer URL in the authorization server metadata does not exactly match the pre-configured issuer URL expected by client. | 2 |
+| **51.5.4** | [ADDED] Verify that the client validates that the ID token is intended to be used for that client (audience) by checking that the 'aud' claim from the token is equal to the 'client_id' value for the client. | 2 |
```
randomstuff, line 115: Let's assume this requirement is not in place in your application A which is using Google (for example) as an OP.

I can register a new application B acting as a OIDC client on Google's OP. A user logins in my application and my server obtains an ID token for this used. Then I could use this ID token to get a session on the user's behalf on application A.

This assumes that A is accepting the implicit or hybrid flows which are accepted to L1 applications.

In think this is pretty bad and is not so much different from not verifying the user password…

I think this should definitely be L1.

elarlang:
@randomstuff - please open those proposal in a separate issue with validating first, that change proposals to L1 are first layer of defense without that you can not be secure (in the case it is the only problem you have)

Issue #2657, opened 2025-03-02
Merge two V14 sections


tghosth:
We currently have the sections [V14.6 Web or Application Server Configuration](https://github.com/OWASP/ASVS/blob/72a27eeb4e3019d2bff257ced1a9f93e78b6ec04/5.0/en/0x22-V14-Config.md#v146-web-or-application-server-configuration) and [V14.7 External Service Configuration](https://github.com/OWASP/ASVS/blob/72a27eeb4e3019d2bff257ced1a9f93e78b6ec04/5.0/en/0x22-V14-Config.md#v147-external-service-configuration)

Practically speaking, the requirements in both are similar and relate to how the application communicates.

I propose that they should be merged.

tghosth:
Given the tag changes, it was easier just to create PR #2659 and we can discuss if further changes are necessary

PR #2658, opened 2025-03-02
Clarify chapter text to resolve #2592

### LINKED ISSUE #2592: Configuration V14 - Sanity Check Comments/Suggestions for v.5.0
**Chapter V14 - Configuration** 

Here are a few things I observed reviewing this chapter  for v.5.0 for sanity check related to #2582 which requires some clarification as maybe my understanding is not enough just by doing a few passes of reading them: 

_Question 1_: For this requirement, `14.1.5 [MODIFIED] Verify that deployed environments are short lived and frequently redeployed to a "known good" but updated state. Alternatively, long lived environments should use some form of "drift prevention" to ensure that deployed configurations are not changed to an insecure state.`,  

What does it mean by "deployed environments", does this pertain to non-production environments or does it mean production environments? Why is there a need for production environments need to be short-lived? Maybe that's one question someone would be struggling with reading this requirement. It seems a bit vague to me. 

_Question 2_: For this requirement, `14.1.9 [ADDED] Verify that application code or functionality can only be changed via the standard update or build process and not directly in production through application functionality or some other direct modification mechanism.`, 

While this is a good common sense control, but upon reading it a few times doesn't this mean through standard build process which means it will deal with CI/CD pipeline? Isn't that also out of scope for ASVS and is treated in another OWASP project I think OWASP SAMM, right?  

_Question 3_: Why did we delete 14.3.1? Old requirement in 4.0.2 says, `14.3.1 [DELETED] Verify that web or application server and framework error messages are configured to deliver user actionable, customized responses to eliminate any unintended security disclosures.`

Isn't this a good idea to prevent application server and framework error messages to be configured against unintended security disclosures? Is it removed because it is not practical to test?

Recommendations: 
* V14.6 Web or Application Server Configuration
Needs some paragraph describing this section. For consistency across all subsections in ASVS. 

* For V14.8 Secret Management, I thought of another control maybe it is already written somewhere, but let me articulate it and let's discuss if we need to add something like it, "Ensure that exposure and access to key material is limited such as a just-in-time access". What do you guys think?

tghosth:
This Pull Request relates to issue #2592

The PR is ready for review, I have marked it as draft to avoid accidental merging.

Text for 14.6 and 14.7 will be handled in issue #2657.

### DIFF
--- a/5.0/en/0x22-V14-Config.md
+++ b/5.0/en/0x22-V14-Config.md
@@ -2,12 +2,11 @@
 
 ## Control Objective
 
-Ensure that a verified application has:
-
-* A secure, repeatable, automatable build environment.
-* Hardened third-party library, dependency, and configuration management to ensure that out-of-date or insecure components are not included in the application.
-
 Configuration of the application out of the box should be safe to be on the Internet, which means a safe out-of-the-box configuration.
+
+This chapter tries to provide guidance on the various configurations that will be necessary to achieve this including both configurations to apply whilst developing the application and also those that are applied at build and deploy time.
+
+This includes topics such as preventing data leakage, securely managing communications between different components and how secrets are protected.
 
 ## V1.14 Configuration Documentation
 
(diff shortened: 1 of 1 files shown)


elarlang (changes requested):
"tries to provide", "tries to capture" - this is like we are not sure does it provide or not

14.3 - section text

I think we should keep this part - it have been discussed too many times.
> For example, hiding the version of server-side components does not fix the need to patch all components, and disabling the folder listing does not eliminate the need to use authorization controls or keep files away from the public folder, but it raises the bar.

tghosth:
Excellent point on "tries". Happy to add back in that other paragraph as well.

elarlang (approved):

PR #2684, opened 2025-03-12
chapter text, #2621

elarlang:
This Pull Request relates to issue #2621

### DIFF
--- a/5.0/en/0x51-V51-OAuth2.md
+++ b/5.0/en/0x51-V51-OAuth2.md
@@ -127,6 +127,8 @@ Note that if using the id-token flow (not the code flow), no access tokens are i
 
 ## V51.7 Consent Management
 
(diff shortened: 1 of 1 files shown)


jmanico (commented):
I've noticed that to format of chapters header vary a lot. Some start with "The following requirements cover" some start with just talking about the topic, some start with "This section covers..."

I think we want more standardization here. I like "This section...".

tghosth (approved):

tghosth:
> I've noticed that to format of chapters header vary a lot. Some start with "The following requirements cover" some start with just talking about the topic, some start with "This section covers..."
> 
> I think we want more standardization here. I like "This section...".

Can you open an issue about this @jmanico ?

jmanico:
Will do.

PR #2719, opened 2025-03-16
Remove out of scope references

tghosth:
Small change to remove references from V1 as they all refer to things that are out of scope for ASVS.


elarlang (approved):

tghosth:
> Note that when I did a cleanup previously, I left them in because maybe there is need to use them in the Recommendation section,

Doh I think you are correct, I opened #2721

Issue #2678, opened 2025-03-06
Mssing requirements related to OAuth/OIDC audience injection attacks?


randomstuff:
See [Notice of a Security Vulnerability](https://openid.net/notice-of-a-security-vulnerability/) for reference about a audience injection attack where an attacker (in this case a malicious AS) might trick a client into issuing a token with an audience claim conflicting with another AS (audience injection attack).

see CVE-2025-27370, CVE-2025-27371

This type of vulnerability might be applicable to other contexts.

Straw-man proposal intended to verify the lack of audience injection vulnerabilities:

> Verify that if a token issuer uses the same private key for issuing tokens to different audiences, the issue tokens contain an audience restriction that uniquely identifies the intended audiences in order to prevent a token to be reused with an unintended audience. If the audience identifier is dynamically provisioned, the token issuer must validate these audiences in order to make sure that they do not result in audience impersonation.

elarlang:
The topic has been raised, and input/update for the proposal comes from Tim Würtele.

elarlang:
So is it a requirement into V52.2?

We also need to check that we don't overlap or duplicate V52.2.3:
> V52.2.3 Verify that the service only accepts tokens which are intended for use with that service (audience). For JWTs, this can be achieved by validating the 'aud' claim against an allowlist defined in the service.

SECtim:
@elarlang Thanks for the credits! I fully agree that overlap with V52.2.3 should be avoided to ensure readers do not mix up the two requirements: Audience injection attacks can _only_ be prevented by the issuer of a token (by selecting an appropriate audience value). The receiver of a token cannot prevent these attacks, no matter what checks they apply to the audience value in the received token.

elarlang:
Well, by this definition I think it is not a perfect match to V52.2 section, as it covers more the step "before using the token content", but not "how to build a token content".

> In any case, we are happy to provide examples/explanations as needed and to review proposed requirements - the one suggested by Elar above already went back and forth with us once.

Here I have been the "gateway" between you (@SECtim) and Gabrierl (@randomstuff). So I just try to use your know-how and organize a valid requirement to a valid section :)

randomstuff:
Yes, I agree with @elarlang and @SECtim: if looks similar to 52.2.3 but is actually different because it is a requirement on the issues whereas 52.2.3 is a requirement about the validator.

We currently mostly have requirements about token consumption/validation with the exception of 52.1.2 which covers both generation and validation.

[...]

PR #2745, opened 2025-03-18
Add requirement about OAuth audience impersonation

randomstuff:
Relates to #2678.

I used L2 as a default. It seems fine for this requirement but feel free to comment on this.


elarlang:
Do we need to touch the section text as well?

randomstuff:
Oh yes, probably. I'll look into that.

5.0/en/0x52-V52-Tokens.md
```diff
@@ -8,15 +8,15 @@ The use of self-contained tokens has become very widespread, even outside of OID
 
 ## V52.1 Token source and integrity
 
-Before inspecting the contents of a self-contained token, it is necessary to ensure that the token has been produced by a trusted party and that it has not been tampered with.
+This section describes requirements about the content of the token both for the issuer and for the consumer of the token in order to prevent unintended consumption of the token (e.g., by the wrong service, for the wrong usage). Before making security decisions based on the content of a self-contained token, the consumer must validate that the token is intended to be consumed by this consumer and was issued for the purpose for which it was presented. On the other hand, the issuer must associate claims to the token in order to make this validation possible and ensure that a malicious actor may not deceive it into including malicious claims in the token.
```
tghosth, line 11: I'm not sure this change is necessary. Some of the text doesn't seem right for this particular section and feels like it repeats what is in 52.2.

I would probably suggest something simple for this section like:

"_This section includes requirements that ensure that the token has been produced by a trusted party and that it has not been tampered with._"

Is there anything else which is super important to add to that?

tghosth (changes requested):

5.0/en/0x52-V52-Tokens.md
elarlang: We need to avoid overlap with V52.2 here. Intended usage feels like V52.2 area.

elarlang: I used proposed short text by Josh and move on at the moment. If there is a need or reason to improve it, we can do it again.

tghosth (approved):

Issue #2201, opened 2024-10-28
Handle Glossary


tghosth:
* Need to check that the glossary is in alphabetical order correctly.
* Need to check if anything is missing.

elarlang:
This is something to do when all requirements are in place:
* check from requirements, that all needed terms are in the glossary
* re-check the glossary, so that there are no un-needed and/or deprecated items
* sync the style for each line
* put glossary lines to correct order

elarlang:
It needs at least minimal validation...

tghosth:
I will try and take this

tghosth:
@oleksiidov rather than capitalisations, if you could help with this then that would actually help a lot

oleksiidov:
@tghosth, will try to help here as much as I can, just needed weekend started ) . And @elarlang provided clear tasks and flow to be done.

oleksiidov:
I've reviewed glossary and created PR #3149 with following updates:
- suggestion on deletion of terms unused in v5 (it seems they came from v4)
- made suggestions on corrections or improvements for some terms (or maybe some of them should be removed also)
- slightly corrected style for some terms to follow unified approach

There following terms could be useful, but not directly mentioned in this version of document, and maybe they also should be deleted:
- Automated Verification
- Design Verification
- Dynamic Verification
- Single-factor authenticator
- Software Composition Analysis
- X.509 Certificate

Colleagues, will be thankful for review PR and comment on how this aligned with what is needed.

If have time, planning to do the following:
- check from requirements description, that all needed terms are in the glossary
- check sorting glossary terms in alphabetical order

tghosth:
This is now sorted correctly, @elarlang is there anything else release blocking that needs to happen here?

oleksiidov:
@tghosth, I have prepared suggestions for adding missing terms (5-8) form chapters to glossary. Do I have time till today EOD to make PR on this?

tghosth:
Yeah sure

PR #2770, opened 2025-03-24
hsts acronym

elarlang:
This Pull Request relates to issue #2773


tghosth:
Should we also update [50.8.4](https://github.com/OWASP/ASVS/blob/f617af09a52d7d19db4ae6a3d17891749c14ea24/5.0/en/0x50-V50-Web-Frontend-Security.md#L102)? Should we add HSTS to the glossary?

elarlang:
I did not recheck that part. You can add, but we have also issue #2201 for that.

tghosth:
@elarlang what do you think about my update?

PR #2776, opened 2025-03-24
Tidy up use of HSTS to resolve #2773

### LINKED ISSUE #2773: HSTS usage in requirements
Related requirements:
* 1.50.1 Verify that application documentation states the expected security features that browsers using the application should support (such as HTTPS, HSTS, Content Security Policy (CSP), and other relevant HTTP security mechanisms). It should also define how the application must behave when some of these features are not available (such as warning the user or blocking access).
* V50.8.4 Verify that the application's top-level domain (e.g., site.tld) is added to the public HSTS preload list so that the use of TLS for the application is built directly into the main browsers, rather than relying only on the relevant HTTP response header field.

"HSTS" is something to write out fully, HTTP Strict Transport Security (HSTS).

For 1.50.1 it is easy, for 50.8.4 it requires some wording change, because "HSTS preload list" is a term of it's own and if to write it as "HTTP Strict Transport Security (HSTS) preload list" may not be helpful.

tghosth:
This Pull Request relates to issue #2773


elarlang (commented):
should vs must?

tghosth:
> should vs must?

Fixed

elarlang (approved):

Issue #2456, opened 2024-12-12
Informative list of tasks and dependencies related to mapping and requirement (re)numbering


elarlang:
Just for your information, what kind of tasks do we.. (I?) have and what are the dependencies between them

* All content of requirements are in - (soft requirement) later it is annoying to track changes
* Requirements texts as style finetuned (soft requirement)
* Levels are changed - required for reordering, that L1 requirements come first
* Prep for keep mapping
  * Mapping related to the v4.0.3 release is exported to a separate file
  * Current requirement number is copied/duplicated to last/additional column for keeping info, where the requirement was located before
* Renumbering everything - remove placeholders and move things around if needed

The rest of the document finetuning

PR #2805, opened 2025-03-28
Asvs 2456 step 2 5 and other corrections

elarlang:
This Pull Request relates to issue #2456


jmanico (approved):

jmanico:
The URL checker is failing on this PR, but it's often faulty. What do you think @elarlang ?

elarlang:
> The URL checker is failing on this PR, but it's often faulty. What do you think @elarlang ?

It failed on checking the URL that actually exists, so it should be temporary

Issue #2513, opened 2025-01-08
Feedback about approved MAC algorithms


randomstuff:
Feedback from Bart Preneel about approved MAC algorithms.

> Add 
> * AES-CMAC
> * AES-GMAC
> * Poly1305-AES

LGTM.

> Remove Blake3 as it is a hash function

Blake3 was added in this list because it [includes a MAC mode](https://www.ietf.org/archive/id/draft-aumasson-blake3-00.html#section-1.1). Maybe this should be clarified as "Blake3 in keyed_hash mode" or something?

tghosth:
@randomstuff, with https://github.com/OWASP/ASVS/pull/2563 merged, is this issue now resolved?

randomstuff:
I 'have added a PR to fix the last comment (about BLAKE3). This issue can be closed once this PR has been merged or rejected :)

elarlang:
But who need to verify or agree with that? @danielcuthbert ?

tghosth:
> But who need to verify or agree with that? [@danielcuthbert](https://github.com/danielcuthbert) ?

Yeah I want to prepare a bunch and then ping him with a todo list :)

randomstuff:
This bas been fixed through #2826 and other.

PR #2826, opened 2025-03-30
Clarify that BLAKE3 has a MAC mode

randomstuff:
Intended to fix the last part of #2513


randomstuff:
Ping  @danielcuthbert.

randomstuff:
Justification for this from the [BLAKE3 paper](https://raw.githubusercontent.com/BLAKE3-team/BLAKE3-specs/master/blake3.pdf):

> Like BLAKE2, BLAKE3 provides a keyed mode, keyed_hash. This removes the need for a
separate construction like HMAC. The keyed_hash mode is also more efficient than keyed
BLAKE2 or HMAC for short messages. BLAKE2 requires an extra compression for the key
block, and HMAC requires three extra compressions. The keyed_hash mode in BLAKE3 does
not require any extra compressions.

elarlang (approved):

Issue #2512, opened 2025-01-08
Feedback about hash functions


randomstuff:
Feedback from Bart Preneel.

> Add a note stating that SHA256 and SHA512 are vulnerable to extension attacks, where an attacker can compute from h(x) and y h(x||y) without knowing x (this is not a property that a well-designed hash function satisfies). 

LGTM.

> Remove KMAC128 and KMAC256 as these are MAC algorithms and not hash functions.

Yes, indeed.

> About Blake2s, Blake2b, Blake3: I just would like to point out that these algorithms are NOT the outcome of an open competition and have received much less scrutiny than other hash functions in the list. 

Does this make sense to add a column in the appendix for this?

randomstuff:
About the first point, possible options:

1. do nothing
2. add a warning in the appendix
3. add a requirement.

I would not choose option 3.

Proposed warning for the appendix (option 2):

> Warning: SHA-1, SHA-256 and SHA-512 are vulnerable to length extension attacks. An attacker can compute the hash h(x || y) from h(x) and y without knowing x.

Would this be enough?

@danielcuthbert, any feedback on this?

danielcuthbert:
I like warning too and we could expand upon this, maybe like:

Warning: SHA-256, and SHA-512 are vulnerable to length extension attacks if used as plain hash functions for authentication or integrity checks. An attacker can compute h(x || y) from h(x) and y without knowing x. To avoid this, use a keyed construction such as HMAC or KMAC when message authentication is required.

I mean if we want to add more clarity, we can add more details about HMAC/KMAC and message auth if you so wish? I've added this into my fix branch

randomstuff:
> I mean if we want to add more clarity, we can add more details about HMAC/KMAC and message auth if you so wish? I've added this into my fix branch

I think we don't want to include too much details. It's probably better to add some reference than going too deep in this topic.

jmanico:
I personally like the more detailed warning, but if you want something shorter maybe just suggest:

 _If you need authenticity or integrity, don’t rely on plain SHA-256/512_

randomstuff:
Oh no I think @danielcuthbert's proposition is good. I was thinking he was asking about adding even more details than this proposition :)

jmanico:
> Oh no I think @danielcuthbert's proposition is good. I was thinking he was asking about adding even more details than this proposition :)

I get it now, I'm with you. I like adding more details here too!

randomstuff:
Done through #2573

PR #2828, opened 2025-03-30
Remove KMAC from lists of hash functions because they are actually MAC

randomstuff:
Addresses part of #2512.

Ping @danielcuthbert, @unprovable


randomstuff:
Already one through #2573

PR #2855, opened 2025-04-03
Update contributing with Elar's mapping explanation

tghosth:
As discussed with @elarlang


tghosth:
Note that the first commit is Elar's original. The 2nd commit is my changes to that to make it easier for @elarlang to see what I did.

Issue #2802, opened 2025-03-27
texts from 3rd person point of view


elarlang:
Except the Preface, ASVS content must not contain "I", "we", "you", etc. Text must be description from 3rd person point of view

from folder `5.0/en/`
```
grep -ri " we " .
grep -ri " you " . | grep -v "WYSIWYG" | grep -v "something you "
```

Front-texts will be rewritten anyway, but in other files it can be changed already.

elarlang:
`0x99-Appendix-X_Recommendations.md`
* Whilst preparing version 5.0 of the Application Security Verification Standard (ASVS), it became clear that there were a number of existing and newly suggested items that we didn't want to include as requirements in 5.0. This may have been because they were not in scope for ASVS as per the definition for 5.0 or alternatively **we** did not feel **we** could make them mandatory, even through we felt they were a good idea.
* **We** did not want to lose these items entirely so we have tried to capture some of them in this appendix.
* The following items are in-scope for ASVS. **We** don't think they should be made mandatory but we would strongly recommend considering them as part of a secure application.

tghosth:
I will try and take this

PR #2887, opened 2025-04-04
AppendixV: add more information about using CBC mode

### LINKED ISSUE #2494: Cryptography - Received comments about CBC
Currently we have this comment the crypto appendix:

> All encrypted messages must be authenticated. Given this, for ANY use of CBC mode there MUST be an associated hashing function or MAC to validate the message. In general, this MUST be applied in the Encrypt-Then-Hash method (but TLS 1.2 uses Hash-Then-Encrypt instead). If this cannot be guaranteed, then CBC MUST NOT be used.

We have received this comment from Bart Preneel:

> I would not hide such an important aspect in an appendix. I would also rewrite it as follows:
>
> > All encrypted messages must be authenticated. Given this, for ANY use of CBC mode there MUST be an associated <del>hashing function or</del> MAC **algorithm** to validate the message. In general, this MUST be applied in the Encrypt-Then-Hash method (but TLS 1.2 uses Hash-Then-Encrypt instead). If this cannot be guaranteed, then CBC MUST NOT be used.
> >
> > **The only application where encryption without a MAC algorithm is allowed is disk encryption.**
>> 
> > **If CBC is used, it shall be guaranteed that the verification of the padding is performed in constant time.**

randomstuff:
Fixes #2494

### DIFF
--- a/5.0/en/0x97-Appendix-V_Cryptography.md
+++ b/5.0/en/0x97-Appendix-V_Cryptography.md
@@ -91,7 +91,8 @@ Approved modes are listed in order of preference.
 
 Notes:
 
-* All encrypted messages must be authenticated. Given this, for ANY use of CBC mode there MUST be an associated hashing function or MAC to validate the message. In general, this MUST be applied in the Encrypt-Then-Hash method (but TLS 1.2 uses Hash-Then-Encrypt instead). If this cannot be guaranteed, then CBC MUST NOT be used.
+* All encrypted messages must be authenticated. For ANY use of CBC mode there MUST be an associated hashing MAC algorithm to validate the message. In general, this MUST be applied in the Encrypt-Then-Hash method (but TLS 1.2 uses Hash-Then-Encrypt instead). If this cannot be guaranteed, then CBC MUST NOT be used. The only application where encryption without a MAC algorithm is allowed is disk encryption.
+* If CBC is used, it shall be guaranteed that the verification of the padding is performed in constant time.
 * When using CCM-8, the MAC tag only has 64 bits of security. This does not conform to requirement 6.2.9 which requires at least 128 bits of security.
 
 ### Key Wrapping


jmanico (commented):
disk encryption is a bit out of our scope in ASVS, suggest removing it.

randomstuff:
> disk encryption is a bit out of our scope in ASVS, suggest removing it.

If we do that, we should remove the lines about XTS, XEX and LRW modes. In this case, I would still include a note saying:

> * Disk encryption is considered out of scope for the ASVS. Therefore this appendix does not list any approved method for disk encryption. For this usage, encryption without authentication is usually accepted and the XTS, XEX and LRW modes are typically used.

jmanico:
> > disk encryption is a bit out of our scope in ASVS, suggest removing it.
> 
> If we do that, we should remove the lines about XTS, XEX and LRW modes. In this case, I would still include a note saying:
> 
> > * Disk encryption is considered out of scope for the ASVS. Therefore this appendix does not list any approved method for disk encryption. For this usage, encryption without authentication is usually accepted and the XTS, XEX and LRW modes are typically used.

I think that is a good solution

randomstuff:
@danielcuthbert et @unprovable, any feedback on that? If not, I'll update the PR accordingly.

randomstuff:
Changes : 

* Move disk encryption modes out of the list of approved modes and into a note
*  Mark CBC as deprecated (addressing #3111)

elarlang:
As you are already there, then a non-technical check based on #2802, and change the "we" form in the "AES Cipher Modes" section.

jmanico (approved):

randomstuff:
@elarlang, I tried to do that and ended up rewriting more of this paragraph.

randomstuff:
I guess this could go into V11.1. I'm not certain this fits very well within "Cryptographic Inventory and Documentation" but that might work.

I would say it is important enough (and different enough from V11.1) to deserve its own requirement.

tghosth:
@randomstuff what is the plan on this PR?

Issue #2889, opened 2025-04-05
V15.4 TOCTUO - change to similar abbreviation


oleksiidov:
Should we follow similar abbreviation for TOCTOU, that is aligned with CWE-367 (https://cwe.mitre.org/data/definitions/367) in [15.4](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x24-V15-Coding.md#v154-concurrency)?

> Concurrency issues such as race conditions, **TOC/TOU** vulnerabilities, deadlocks, livelocks, thread starvation, and improper synchronization can lead to unpredictable behavior and security risks. This section includes various techniques and strategies to help mitigate these risks.
Here we have TOC/TOU

In [15.4.3](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x24-V15-Coding.md#v154-concurrency) we have
> Verify that all access to shared resources is consistently checked and accessed in a single atomic operation to prevent **Time-of-Check to Time-of-Use (TOCTOU)** race conditions, ensuring resource state consistency between check and use. 

Suggestion: 
In start of section put following: 
> Concurrency issues such as race conditions,  **Time-of-Check to Time-of-Use (TOCTOU)** vulnerabilities, deadlocks, ....

And add this abbreviation to [Appendix A](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x90-Appendix-A_Glossary.md)

elarlang:
Also related to #1875

oleksiidov:
@elarlang, I meant to remove "/" from "TOC/TOU", and use just "TOCTOU" everywhere.

elarlang:
Ack, all those problems should be solved anyway:
* one way and style for acronym (TOCTOU)
* one way and style for presenting it in text (most likely lowercase "time-of-check to time-of-use as it is not an official technical name for anything) - depends on #1875 
* add it to the Glossary - #2201

tghosth:
Sounds like specifically for this issue, we should unify use of TOCTOU and add it to the glossary. Should @oleksiidov open a PR with that?

oleksiidov:
@tghosth, @elarlang, thank you for comments! 
Yes I will open PR for this.

elarlang:
This one is merged.

PR #2906, opened 2025-04-06
V15.4 change to similar TOCTUO abbreviation

oleksiidov:
This Pull Request relates to issue #2889


elarlang (changes requested):
Some background "requirements" for requirements - requirements are used out of this document context as a stand-alone piece of text, which means it is not enough when the acronym is explained in the document already, it should also be explained in the requirement itself. So please write it in full length also into requirement 15.4.3.

In the glossary we should not use links. Although I saw that we have one there.

elarlang (approved):

elarlang:
I'll leave it to @tghosth  to review the Glossary row.

oleksiidov:
@elarlang thank you for comments! Will consider this in the future suggestions.

tghosth:
@oleksiidov I made some changes, if you disagree with them, please open up a separate PR.

tghosth:
Thanks for your PR!

oleksiidov:
@tghosth, thank you for changing. Really more close to ASVS purpose.
Slightly changed and created (PR2909)[https://github.com/OWASP/ASVS/pull/2909].

Issue #2987, opened 2025-04-18
11.6.2 / v5.0.be-6.7.1 should be reduced to level 2


ingo-rauner:
Control [11.6.2](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x20-V11-Cryptography.md#v116-public-key-cryptography) is also required for applications with moderate security.

The high security requirements can be addressed with a new control, see issue #2986.

jmanico:
I support this being bumped to a level 2.

CC @randomstuff @danielcuthbert 

The current text is:

_11.6.2 Verify that industry-proven cryptographic algorithms are used for key exchange (such as Diffie-Hellman) with a focus on ensuring that key exchange mechanisms use secure parameters. This will prevent attacks on the key establishment process which could lead to adversary-in-the-middle attacks or cryptographic breaks._

randomstuff:
I totally agree that it should be L2.

jmanico:
Hey @danielcuthbert @elarlang @tghosth are we ok to go to PR on this one?

tghosth:
See discussion here:
https://github.com/OWASP/ASVS/issues/2695

The tl;dr is that the TLS aspects of key exchange are handled in a different chapter and therefore the actual applicability of this requirement is very niche. We have somewhere between 150 and 200 requirements in L2 and we need to make tricky but necessary decisions about what gets pushed up to L3 which is what we did here.

jmanico:
> See discussion here: [#2695](https://github.com/OWASP/ASVS/issues/2695)
> 
> The tl;dr is that the TLS aspects of key exchange are handled in a different chapter and therefore the actual applicability of this requirement is very niche. We have somewhere between 150 and 200 requirements in L2 and we need to make tricky but necessary decisions about what gets pushed up to L3 which is what we did here.

Josh, there was very little discussion on this exact issue in the thread you cited. This is as good as any a place to discuss it. It's a very small change, approved by one of the main cryptographic leads. This is a small change that three of us now agree with. Please consider differently.

tghosth:
@jmanico there was a fair amount of back and forth specifically about this requirement (formerly 6.7.1) and I don't think it is fair to dismiss my request that that is referenced. In fact, you even suggested merging this requirement into a larger TLS requirement.

[...]

PR #2994, opened 2025-04-19
Issue 2987

jmanico:
Addressing https://github.com/OWASP/ASVS/issues/2987


tghosth:
Let's wait for issue resolution

elarlang:
Requested review, but I don't see agreement here: https://github.com/OWASP/ASVS/issues/2987#issuecomment-2816902971

jmanico:
We have three yes to this and one no. I agree we do not have concensus on this yet.

elarlang:
Note, that it is also not about voting. It is aligning with the rest of the document and balancing the levels.

jmanico:
It's very much about voting. There is no one leader to ASVS. There is a group. No one person makes the call.

tghosth:
I am going to close this as I think the discussion went in a different direction so I want to leave @randomstuff space to make the wording clarifications he was [discussing here](https://github.com/OWASP/ASVS/issues/2987#issuecomment-2817977721).

PR #3029, opened 2025-04-22
Update 0x10-V1-Encoding-Sanitization.md

drveresh:
Additional key considerations for Validation and Sanitization Architecture.


tghosth:
Hi @drveresh, thanks for your input :)

The convention for larger changes to ASVS is to start with an issue and then move to a PR once there is a agreement.

There are some more details here:
https://github.com/OWASP/ASVS/blob/master/CONTRIBUTING.md#how-can-i-help

When you open an issue for new requirements, please consider the following:

- Have you verified that no similar requirement exists already in the latest version of the standard? 
- Have you searched for related issues where this might have already been discussed?
- Is the application less secure if the requirement is not in place?
- Is the new requirement in scope for ASVS, broadly speaking is it something in the developer's control?
- Is the new requirement testable?
- is the new requirement implementable by a development team?

Thanks!

PR #3030, opened 2025-04-22
Change suggestions for v3 chapter text

jmanico:
v3 has a great deal of text that is lighting up my grammar checker. This is a major suggestion that I do not expect at all to go live is. And my grammar checkers are lighting up so much that I'm going to hold off on this until closer to release.

But there are many very good suggestions here worth reviewing.

### DIFF
--- a/5.0/en/0x12-V3-Web-Frontend-Security.md
+++ b/5.0/en/0x12-V3-Web-Frontend-Security.md
@@ -2,69 +2,69 @@
 
 ## Control Objective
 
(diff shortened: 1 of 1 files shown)


elarlang:
Those are not grammar changes, those are quite some rewriting wording for V3.4, V3.5, V3.7 requirements.

For example, correct term is HTTP header field. HSTS must be written out.

In short, if the wording of the requirement is changed, it must discussed and be agreed. Those are not just grammar changes.

jmanico:
Elar, I updated the PR to only address minor changes in chapter text. Do any of these work for you?

elarlang (approved):

Issue #2970, opened 2025-04-15
L3 MFA: Are we mandating "supported" or "required


sydseter:
In the 6.3 text we say:

> This section contains general requirements for the security of authentication mechanisms as well as setting out the different expectations for levels including requiring multi-factor for L2 and supporting hardware-based mechanisms for L3, such as standards provided by the Fast IDentity Online (FIDO) Alliance.

All of the larger IDPs supports hardware-based authentication mechanisms. Simply supporting hardware-based authentication mechanisms does not mean that the users of such authentication systems are more secure. I therefor see little reason why you should make a point of it for L3 unless you **require** the use of hardware-based authentication, that is, unless a clear plan accompanied an exception on how the risks around authentication will be mitigated, taking into account NIST's guidance and research on the topic.. I would rather see that ASVS 5  Level 3 aligned with AAL3 in this regard. 

Please see comment from @tghosth https://github.com/OWASP/ASVS/issues/2925#issuecomment-2792342067

> Clarify in the requirement whether single-factor hardware-based auth is sufficient for L3 or whether multi-factor is still required

Please read: https://pages.nist.gov/800-63-3-Implementation-Resources/63B/AAL/

> Although two authentication factors are required at AAL3, one combination of authenticators (Hardware Single-Factor OTP Device plus a Single-Factor Cryptographic Software Authenticator plus a Memorized Secret) consists of three authenticators. This combination stems from the fact that the hardware-based Single-Factor OTP Devices do not provide verifier impersonation resistance, so a Single-Factor Cryptographic Software Authenticator can satisfy that requirement. But since both of those authenticators are something you have, a Memorized Secret is required to satisfy the requirement for two different authentication factors.

This clarifies the need to require MFA for Level 3, but it goes beyond that listing, specifically the 3 requirement necessary to satisfy AAL3.

In addition to all requirements inherent to AAL2, the following requirements apply to AAL3, to my understanding:

- The use of authenticators must guarantee impersonation resistance
- The use of authenticators must guarantee compromise resistance
- The use of authenticators must guarantee authentication intent

In the interest of aligning ASVS with NIST, it seems reasonable to ensure that these 3 requirements are needed in order to ensure that an application satisfy ASVS Level 3.

This is consistent with NIST SP800 63b, 2.3.1: https://pages.nist.gov/800-63-4/sp800-63b/aal/#aal3 

> AAL3 authentication SHALL require one of the following authenticator combinations:
> 
[...]

PR #3081, opened 2025-05-04
Resolve #2925 by clarifying MFA enforcement at L2

### LINKED ISSUE #2925: L2 MFA: Are we mandating "supported" or "required"
In the 6.3 text we say:

> This section contains general requirements for the security of authentication mechanisms as well as setting out the different expectations for levels including requiring multi-factor for L2 and supporting hardware-based mechanisms for L3, such as standards provided by the Fast IDentity Online (FIDO) Alliance.

I think we need to clarify something fundamental here as I don't think we have properly done so up until now.

Which of the following two things are we saying here:

1) Your application must support MFA
2) You cannot allow users to use your application unless they have enabled MFA

I think we have been thinking along the lines of Option 2 whereas that seems completely unrealistic for L2. Maybe more realistic for L3.

If we assume that Option 1 is the correct interpretation then that changes how we consider the password requirements as password security will still be highly relevant at L2.

What do we think @elarlang @jmanico  @randomstuff @ryarmst ?

tghosth:
This Pull Request relates to issue #2925

### DIFF
--- a/5.0/en/0x15-V6-Authentication.md
+++ b/5.0/en/0x15-V6-Authentication.md
@@ -45,7 +45,9 @@ The requirements in this section mostly relate to [&sect; 5.1.1.2](https://pages
 
 ## V6.3 General Authentication Security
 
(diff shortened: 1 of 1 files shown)


elarlang (approved):
Approving L2 changes.

L3-related section text is to be decided via #2970

PR #3085, opened 2025-05-04
Link to 6.8 to resolve #2934

### LINKED ISSUE #2934: 10.5.5 to more generic (non-OAuth specific)
Discussion to make 10.5.5 more generic requirement for Authentication section.

---

edit: from placeholder to actual issue.

| **10.5.5** | Verify that, if specific authentication strength, methods, or recentness is required, the relying party (RP) verifies that the presented ID Token satisfies these constraints (using the 'acr', 'amr' and 'auth_time' claims respectively). | 2 | v5.0.be-51.5.5 |

Initial goal for the issue: add abstract requirement to require require authentication strength, method or recentness from IdP.

Requirement should belong to section "V6.8 Authentication with an Identity Provider"

---

Additional goal comes from to issue: #2925 and with a problem statement, that IdP's may not provide or support such information change.

From @sydseter in https://github.com/OWASP/ASVS/issues/2925#issuecomment-2789105581
> Important to note that enforcement is not possible on the app side, today, as the amr claim not is available to apps using IDPs like Entra ID or Google Identity Platform. Should that be ignored? Given how long the next version of ASVS will live, I believe it should.

My proposal (https://github.com/OWASP/ASVS/issues/2925#issuecomment-2789160831) for additional requirement or "problem to solve" with the initial requirement:
> I don't think we should ignore it, but ask something like "if the strength or method of authentication is not verifiable from IdP, then it must be considered as single-factor and the application must require second factor verification". For example, authenticating the user via Gmail and then asking additionally TOTP via Google Authenticator.

tghosth:
This Pull Request relates to issue #2934

### DIFF
--- a/5.0/en/0x19-V10-OAuth2-and-OIDC.md
+++ b/5.0/en/0x19-V10-OAuth2-and-OIDC.md
@@ -107,6 +107,8 @@ For client authentication, the 'self_signed_tls_client_auth' method is allowed w
 
 As the OIDC relying party acts as an OAuth client, the requirements from the section "OAuth Client" apply as well.
 
+Note that the "Authentication with an Identity Provider" section in the "Authentication" chapter also contains requirements that are relevant but not OIDC/OAuth specific.
+
 | # | Description | Level | #v5.0.be |
 | :---: | :--- | :---: | :---: |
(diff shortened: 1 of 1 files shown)


tghosth:
Draft to prevent accidental merge

elarlang (approved):

PR #3093, opened 2025-05-06
Resolve #2970 by clarifying hardware-based enforcement at L3

### LINKED ISSUE #2970: L3 MFA: Are we mandating "supported" or "required

sydseter:
This Pull Request relates to issue #2970 

To sum it up...

The changes here reflect that we have little time before the ASVS 5 release, now in May. So, in the interest of speed and to avoid ending up in a situation where L3 goes from being supported to being required after the release of ASVS 5 I am suggesting only adding changes to 6.3 for now.

Instead of:

> supporting hardware-based mechanisms (such as standards provided by the Fast IDentity Online (FIDO) Alliance)

I am suggesting:

_hardware-based authentication, performed in an attested and trusted execution environment (TEE), is required for L3_

I have replaced: 

> (such as WebAuthn)

with:

_(such as device-bound passkeys, eIDAS LoA3 enforced authenticators, or authenticators with AAL3 assurance, or the equivalent thereof)_

"or the equivalent" is vital because each European member nation is implementing their eIDAS LoA3 authenticator that is not necessarily based on FIDO2 or WebAuthN
Otherwise, you would disfranchise AAL3 authenticators that not necessarily fits in with what NIST currently has stated (authenticators in UK and countries outside EU, but part of the inner market comes to mind). ref: https://ec.europa.eu/digital-building-blocks/sites/display/EUDIGITALIDENTITYWALLET/EU+Digital+Identity+Wallet+Home

As an alternative, it’s possible to shorten it to:

_(such as authenticators that provides AAL3- or eIDAS LoA3 assurance, or the equivalent thereof)_

eIDAs LoA3 stands side-by-side with AAL3 as NIST considers them to give equivalent assurances. See the "Approach to assurance level mapping" on page 16 for mapping reference between AAL and eID https://www.nist.gov/system/files/documents/2023/12/22/EU-US%20TTC%20WG1_Digital_Identity_Mapping_Report_Final%20Draft%20for%20Comment_22122023.pdf

The rest is just rearranging the sentence to make it flow.

The reason for replacing FIDO and WebAuthn is not that it isn't relevant; it's just that it's not precise enough.  WebAuthn is a web standard, and FIDO is a technical specification describing both AAL2 and AAL3 authenticators.

I am mentioning attestation and TEE because they align with the true definition of what an AAL3 authenticator is supposed to be.

E.g:

«Confidential Computing is the protection of data in use by performing computation in a hardware-based, attested Trusted Execution Environment."», see definition under 2.1: https://confidentialcomputing.io/wp-content/uploads/sites/10/2023/03/CCC-A-Technical-Analysis-of-Confidential-Computing-v1.3_unlocked.pdf

And as we all agree now, AAL3 must be required for L3.


elarlang:
Marked as draft as I don't see an agreement in the related issue. The change should be discussed and agreed on in the issue first, otherwise it's later complicated to follow the rationale and reasoning for the changes, if the discussion is kept in multiple places or the discussion is done over changes in a PR.

elarlang:
There is no agreement in the related issue - there is no need to mark it as ready. It is also going into different direction than proposed and discussed in #3094. In short - there is no need to open a PR if there is no agreement in the issue to do so.

elarlang:
Requirements addressed in this PR will be merged via #3101 so this PR is going to be closed.

Issue #3094, opened 2025-05-06
Discussion: merge L2 and L3 MFA requirements (6.3.3, 6.3.8) and move to section V6.5


elarlang:
This issue is mostly focusing on organising MFA-related requirements. The key aspects for content should be agreed in #2970.

We have 2 MFA-related requirements in section "V6.3 General Authentication Security" but we also have a dedicated MFA-related section "V6.5 General Multi-factor authentication requirements"

| # | Description | Level | #v5.0.be |
| :---: | :--- | :---: | :---: |
| **6.3.3** | Verify that either a multi-factor authentication mechanism or a combination of single-factor authentication mechanisms, must be used in order to access the application. Relaxing this requirement requires a fully documented rationale and a comprehensive set of mitigating controls. | 2 | v5.0.be-2.2.9 |
| **6.3.8** | Verify that a hardware-based authentication mechanism is supported that provides impersonation resistance against phishing attacks (such as WebAuthn) and verifies intent to authenticate by requiring a user-initiated action (such as a button press on a FIDO hardware key). | 3 | v5.0.be-2.2.4 |

Problem to solve 1 (must have): 6.3.3 and 6.3.8 are in conflict from the "accepted for L3" point of view (6.3.3 is accepted for L2 and L3, and 6.3.8 is accepted for L3). So I think we should merge it into one requirement with an in-requirement level description.

Problem to solve 2 (nice to have): We have a dedicated section for MFA, and it would be more logical to search MFA-related requirements from that section.

Additional (nice to have) topic: I would prioritize less password-related requirements and move them down in chapter order.

tghosth:
> Problem to solve 1 (must have): 6.3.3 and 6.3.8 are in conflict from the "accepted for L3" point of view (6.3.3 is accepted for L2 and L3, and 6.3.8 is accepted for L3). So I think we should merge it into one requirement with an in-requirement level description.

Why is this currently a problem. Are you preempting a situation where L3 changes to "required" rather than "supported"?

> Problem to solve 2 (nice to have): We have a dedicated section for MFA, and it would be more logical to search MFA-related requirements from that section.

Agree that makes sense.

> Additional (nice to have) topic: I would prioritize less password-related requirements and move them down in chapter order.

Not sure I would bother, these concerns are still relevant, especially at L1.

elarlang:
[...]

PR #3101, opened 2025-05-08
one mfa (6.3.3+6.3.8) requirement to cover l2 and l3, #3094

elarlang:
This Pull Request relates to issue #3094


tghosth:
I updated section text as well, I have approved for now so you can squash and merge if you are ok with that

PR #3116, opened 2025-05-15
Clarify and focus concurrency reqs to resolve #2960

### LINKED ISSUE #2960: V15.4 "Concurrency" requirements
Current requirements:
| **15.4.1** | Verify that only thread-safe types are used in multi-threaded contexts, or that non-thread-safe types are synchronized to prevent race conditions. | 3 | v5.0.be-10.7.1 |
| **15.4.2** | Verify that concurrent access to shared resources is controlled using synchronization primitives (e.g., locks, mutexes, semaphores) to prevent race conditions and ensure atomic operations on these resources. | 3 | v5.0.be-10.7.2 |
| **15.4.3** | Verify that all access to shared resources is consistently checked and accessed in a single atomic operation to prevent time-of-check to time-of-use (TOCTOU) race conditions, ensuring resource state consistency between check and use. | 3 | v5.0.be-10.7.3 |
| **15.4.4** | Verify that resource acquisition uses a consistent locking strategy to avoid circular dependencies and ensure forward progress, preventing both deadlocks and livelock scenarios. | 3 | v5.0.be-10.7.4 |
| **15.4.5** | Verify that resource allocation policies prevent thread starvation by ensuring fair access to resources, such as by leveraging thread pools, allowing lower-priority threads to proceed within a reasonable timeframe. | 3 | v5.0.be-10.7.5 |
| **15.4.6** | Verify that locking primitives are only accessible to the owning class or module and are not publicly modifiable, ensuring that locks cannot be inadvertently or maliciously modified by external classes or code. | 3 | v5.0.be-10.7.6 |

We have 6 "feels similar" requirements that found it's way into the repo during Summit. Follow-up for issue #2231 to be sure that we can distinguish the requirements from each other and re-check potential overlap with other requirements.

tghosth:
This Pull Request relates to issue #2960

### DIFF
--- a/5.0/en/0x24-V15-Coding.md
+++ b/5.0/en/0x24-V15-Coding.md
@@ -60,12 +60,10 @@ Concurrency issues such as race conditions, time-of-check to time-of-use (TOCTOU
 
 | # | Description | Level | #v5.0.be |
 | :---: | :--- | :---: | :---: |
-| **15.4.1** | Verify that only thread-safe types are used in multi-threaded contexts, or that non-thread-safe types are synchronized to prevent race conditions. | 3 | v5.0.be-10.7.1 |
-| **15.4.2** | Verify that concurrent access to shared resources is controlled using synchronization primitives (e.g., locks, mutexes, semaphores) to prevent race conditions and ensure atomic operations on these resources. | 3 | v5.0.be-10.7.2 |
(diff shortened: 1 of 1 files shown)


tghosth:
@elar will need to consider mappings.

elarlang (approved):

PR #3122, opened 2025-05-19
Clarifications to resolve #3109

### LINKED ISSUE #3109: Clarify why 15.3.1/v5.0.be-10.4.5 is not access control
This requirement got added as a result of #934

| **15.3.1** | Verify that the application only returns data which the user has permission to access. For example, the API response does not return a full object with attributes that contain values the user has no permission to access, despite having permission to access the data object itself. | 1 | v5.0.be-10.4.5 |

If I understand correctly, the reason it is here is because it is talking about data that *no* user should have access to and therefore it is not a question of different permissions for different users.

If this is the case, I think we should refine as follows:

| **15.3.1** | Verify that the application only returns data which users should be able to access. For example, the API response does not return a full object which includes data attributes that no user should be able to see, despite having permission to access the data object itself. | 1 | v5.0.be-10.4.5 |

What do you think @elarlang ?

tghosth:
This Pull Request relates to issue #3109

Draft to prevent accidental merging

### DIFF
--- a/5.0/en/0x17-V8-Authorization.md
+++ b/5.0/en/0x17-V8-Authorization.md
@@ -14,7 +14,7 @@ Comprehensive authorization documentation is essential to ensure that security d
 | # | Description | Level | #v5.0.be |
 | :---: | :--- | :---: | :---: |
 | **8.1.1** | Verify that authorization documentation defines rules for restricting function-level and data-specific access based on consumer permissions and resource attributes. | 1 | v5.0.be-1.4.7 |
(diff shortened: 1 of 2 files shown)


elarlang (commented):
8.1.2
* "Note that these rules might depend on the state of the relevant data object" - not sure what it helps to explain. The state here should be taken as an example, "Note that these rules might depend on the attribute value of the relevant data object, such as state or status."

15.3.1
* "via application" is not needed?
* "no user should be able to see" is commented here, https://github.com/OWASP/ASVS/issues/3109#issuecomment-2870026278. I think it is not correct way to go.

tghosth:
Merge conflicted myself...

PR #3126, opened 2025-05-21
Clarifications to resolve #3109

### LINKED ISSUE #3109: Clarify why 15.3.1/v5.0.be-10.4.5 is not access control

tghosth:
This Pull Request relates to issue #3109

This PR replaces the broken #3122

### DIFF
--- a/5.0/en/0x17-V8-Authorization.md
+++ b/5.0/en/0x17-V8-Authorization.md
@@ -14,7 +14,7 @@ Comprehensive authorization documentation is essential to ensure that security d
 | # | Description | Level | #v5.0.be |
 | :---: | :--- | :---: | :---: |
 | **8.1.1** | Verify that authorization documentation defines rules for restricting function-level and data-specific access based on consumer permissions and resource attributes. | 1 | v5.0.be-1.4.7 |
-| **8.1.2** | Verify that authorization documentation defines rules for field-level access restrictions based on consumer permissions and resource attributes. | 2 | v5.0.be-1.4.8 |
+| **8.1.2** | Verify that authorization documentation defines rules for field-level access restrictions (both read and write) based on consumer permissions and resource attributes. Note that these rules might depend on the state of the relevant data object. | 2 | v5.0.be-1.4.8 |
 | **8.1.3** | Verify that the application's documentation defines the environmental and contextual attributes (including but not limited to, time of day, user location, IP address, or device) that are used in the application to make security decisions, including those pertaining to authentication and authorization. | 3 | v5.0.be-1.4.6 |
 | **8.1.4** | Verify that authentication and authorization documentation defines how environmental and contextual factors are used in decision-making, in addition to function-level, data-specific, and field-level authorization. This should include the attributes evaluated, thresholds for risk, and actions taken (e.g., allow, challenge, deny, step-up authentication). | 3 | v5.0.be-1.4.9 |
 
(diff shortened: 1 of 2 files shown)


tghosth:
@elarlang had comments there:

> 8.1.2
> * "Note that these rules might depend on the state of the relevant data object" - not sure what it helps to explain. The state here should be taken as an example, "Note that these rules might depend on the attribute value of the relevant data object, such as state or status."
> 
> 15.3.1
> * "via application" is not needed?
> * "no user should be able to see" is commented here, https://github.com/OWASP/ASVS/issues/3109#issuecomment-2870026278. I think it is not correct way to go.

tghosth:
> 8.1.2
> * "Note that these rules might depend on the state of the relevant data object" - not sure what it helps to explain. The state here should be taken as an example, "Note that these rules might depend on the attribute value of the relevant data object, such as state or status."

Updated the wording.
 
> 15.3.1
> * "via application" is not needed?

Ok

> * "no user should be able to see" is commented here, https://github.com/OWASP/ASVS/issues/3109#issuecomment-2870026278. I think it is not correct way to go.

This is the crux of this point and I have [explained this already](https://github.com/OWASP/ASVS/issues/3109#issuecomment-2870028084). 

This requirement should be talking about data that no user should ever be able to access via the application, with the example that it is the opposite of mass assignment. That is why it is worded like this.

elarlang:
> This requirement should be talking about data that no user should ever be able to access via the application, with the example that it is the opposite of mass assignment. That is why it is worded like this.

I don't agree with that. If an admin user has permission to see it somewhere else, it does not make it valid for this current user to see those attributes. I think it is misleading and makes the requirement less verifiable.

tghosth:
> > This requirement should be talking about data that no user should ever be able to access via the application, with the example that it is the opposite of mass assignment. That is why it is worded like this.
> 
> I don't agree with that. If an admin user has permission to see it somewhere else, it does not make it valid for this current user to see those attributes. I think it is misleading and makes the requirement less verifiable.

In which case we need to remove this requirement as it is covered by 8.1.2

elarlang:
> > > This requirement should be talking about data that no user should ever be able to access via the application, with the example that it is the opposite of mass assignment. That is why it is worded like this.
> > 
> > 
> > I don't agree with that. If an admin user has permission to see it somewhere else, it does not make it valid for this current user to see those attributes. I think it is misleading and makes the requirement less verifiable.
> 
> In which case we need to remove this requirement as it is covered by 8.1.2

I don't agree with that neither, https://github.com/OWASP/ASVS/issues/3109#issuecomment-2870966568

tghosth:
> > > > This requirement should be talking about data that no user should ever be able to access via the application, with the example that it is the opposite of mass assignment. That is why it is worded like this.
> > > 
> > > 
> > > I don't agree with that. If an admin user has permission to see it somewhere else, it does not make it valid for this current user to see those attributes. I think it is misleading and makes the requirement less verifiable.
> > 
> > 
> > In which case we need to remove this requirement as it is covered by 8.1.2
> 
> I don't agree with that neither, [#3109 (comment)](https://github.com/OWASP/ASVS/issues/3109#issuecomment-2870966568)

I mean, you said yourself:

> So, probably it is not wrong to move it to the authorization chapter and merge it to 8.2.3 (but it is L2, compared to 15.3.1 L1)

I have read that comment multiple times and I don't understand why it is not field-level access control, we even clarified the relevant requirement to be clear it is about both read and write

elarlang (approved):

PR #3127, opened 2025-05-21
Add clarifications on key sharing to resolve #2310

### LINKED ISSUE #2310: V6 - Proper MAC (in contrast to digital signatures)
Should there be a requirement about the proper usage of MAC (in contrast to digital signatures). In particular, if there are more than two participants, MAC is usually not safe.

See for example [Differences between Digital Signatures and MACs](https://datatracker.ietf.org/doc/html/rfc7515#section-10.5) in the JWS RFC (emphasis mine):

> Both signatures and MACs provide for integrity checking -- verifying that the message has not been modified since the integrity value was computed.  However, MACs provide for origination identification only under specific circumstances.  It can normally be assumed that a private key used for a signature is only in the hands of a single entity (although perhaps a distributed entity, in the case of replicated servers); however, a MAC key needs to be in the hands of all the entities that use it for integrity computation and checking. Validation of a MAC only provides corroboration that the message was generated by one of the parties that knows the symmetric MAC key. This means that <mark>origination can only be determined if a MAC key is known only to two entities and the recipient knows that it did not create the message</mark>.  MAC validation cannot be used to prove origination to a third party.

Note: another problematic scenario mentioned here is when a MAC-ed message send from A to B could be sent again from B to A.

tghosth:
This Pull Request relates to issue #2310

### DIFF
--- a/5.0/en/0x20-V11-Cryptography.md
+++ b/5.0/en/0x20-V11-Cryptography.md
@@ -26,7 +26,7 @@ Please see the Appendix for current guidance on vetted PQC primitives and standa
 
 | # | Description | Level | #v5.0.be |
 | :---: | :--- | :---: | :---: |
-| **11.1.1** | Verify that there is a documented policy for management of cryptographic keys and a cryptographic key lifecycle that follows a key management standard such as NIST SP 800-57. This should include ensuring that keys are not overshared (for example with more than two entities for shared secrets and more than entity for private keys.) | 2 | v5.0.be-1.6.1 |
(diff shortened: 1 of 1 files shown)


elarlang (approved):

elarlang:
Review my changes and merge if agreed

tghosth:
> Review my changes and merge if agreed
> 
> The main concern is:
> 
> > for example, with more than two entities for shared secrets and more than **one** entity for private keys
> 
> I added the "one" there, but maybe some correct value was missing from there?

Nah that looks correct, thanks!

PR #3129, opened 2025-05-21
Update preface

tghosth:
This Pull Request relates to issue ????

This takes the draft from the Google doc and refines the writing style whilst maintaining the core message. The first commit

### DIFF
--- a/5.0/en/0x02-Preface.md
+++ b/5.0/en/0x02-Preface.md
@@ -1,57 +1,23 @@
 # Preface
 
-Welcome to the Application Security Verification Standard (ASVS) version 5.0!
-
-## Introduction
+Welcome to the Application Security Verification Standard (ASVS) version 5.0.
 
 Established in 2008 through a collaborative community effort, the ASVS defines the security requirements to be considered when designing, developing, and testing modern web applications and web services.
 
+The previous major version (4.0) was released in 2019 with the most recent minor version released in 2021 (4.0.3) so it is time for a new major version, updated and adapted with the latest developments in software security. 
+
 ASVS version 5.0 is the culmination of significant effort from the leaders, the working group, and other community members to update and improve this important standard.
-
-Our goal for this version has been to make the ASVS easier to use while also making it more clearly focused on its defined scope and covering new, important areas of application development.
-
-## Key Objectives from ASVS Version 5.0
 
 Version 5.0 has been developed with several key principles in mind.
 
-### Clear Set of Requirements
+The scope is defined more precisely to follow the words Application, Security, Verification, and Standard. The redefined scope and focus for requirements have changed the direction for wording to move towards preventing security problems instead of asking to implement a certain technical solution. Requirement texts also got attention to make them self-explanatory as to why the requirement exists.
 
-The requirements for version 5.0 were prepared based on the following considerations:
+Concept of documented security decisions as requirements is included in the ASVS to provide required information for implementing and verifying implementation requirements, and also to provide a flexibility mechanism for organizations, as each organization and application may have their own specific needs and risks to mitigate.
 
-* Aggressively deduplicate requirements to avoid controls or concepts being split across multiple places.
-* Clarify unclear or non-actionable requirement text.
-* Add new requirements to cover areas of particular concern, such as permissions, tokens, and cryptography.
-* Introduce new chapters and sections for areas that might not apply to all applications yet remain security-sensitive (e.g., OAuth and WebSockets).
+The philosophy for requirements levels in version 5.0 has changed, and levels are designed to make the ASVS easier to adopt. It still has three levels, starting from Level 1, which is built mainly for the requirements that are the first layer of defense, and to take the first step into ASVS. Level 2 can be considered as "standard security", and Level 3 as an "extra step forward".
 
-### Clarifying the Scope of the Standard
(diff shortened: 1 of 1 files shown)


5.0/en/0x02-Preface.md
```diff
@@ -1,57 +1,29 @@
...
+ASVS 5.0 is the result of extensive contributions from project leaders, working group members, and the wider OWASP community to update and improve this important standard.
 
-## Key Objectives from ASVS Version 5.0
+## Principles behind version 5.0
 
-Version 5.0 has been developed with several key principles in mind.
+This major revision has been developed with several key principles in mind:
 
-### Clear Set of Requirements
+* Refined Scope and Focus: The standard now aligns more directly with the foundational pillars in its name: Application, Security, Verification, and Standard. Requirements have been rewritten to emphasize the prevention of security flaws rather than mandating specific technical implementations. Each requirement is now crafted to be self-explanatory, highlighting its rationale.
```
elarlang, line 17: 4 times "now" in this list

> The standard now aligns more directly

"The standard is made to align more directly"

---

> Each requirement is now crafted to be self-explanatory, highlighting its rationale.

It is a direction, can not state the statement is satisfied.

5.0/en/0x02-Preface.md
```diff
@@ -1,57 +1,29 @@
...
 
-The requirements for version 5.0 were prepared based on the following considerations:
+* Support for Documented Security Decisions: ASVS 5.0 introduces requirements for documenting key security decisions. This enhances traceability and supports context-sensitive implementations, allowing organizations to tailor their security posture to their specific specific needs and risks.
 
-* Aggressively deduplicate requirements to avoid controls or concepts being split across multiple places.
-* Clarify unclear or non-actionable requirement text.
-* Add new requirements to cover areas of particular concern, such as permissions, tokens, and cryptography.
-* Introduce new chapters and sections for areas that might not apply to all applications yet remain security-sensitive (e.g., OAuth and WebSockets).
+* Updated Levels: While ASVS retains its three-tier model, the level definitions have evolved to make the ASVS easier to adopt. Level 1 now serves as a foundational entry point focused on primary defenses. Level 2 represents a comprehensive viewo of standard security practices, and Level 3 addresses advanced, high-assurance requirements.
```
elarlang, line 21: would like to use "1st layer of defence" for Level 1, as this is the term we/I use everywhere for L1.

Remove "now" from "Level 1 now serves"

5.0/en/0x02-Preface.md
```diff
@@ -1,57 +1,29 @@
...
-It is important that all requirements are relevant to the defined scope of the standard and are worded consistently with its goals. The guidelines for this were:
+## Looking ahead
 
-* Ensure that all requirements are within the scope of a web application or service.
-* Verify that requirements align with the ASVS name, specifically:
-    * **Application** – Requirements are at the application level and are the responsibility of application developers.
-    * **Security** – Requirements are clearly necessary to secure the application.
-    * **Verification** – Requirements are written with a clear, verifiable goal.
-    * **Standard** – Requirements exhibit clear consistency and structure, as expected from a standard.
+The ASVS is not a static document. Although Version 5.0 is a major release, development continues. This release allows the wider community to benefit from the improvements and additions which have been accumulated but also lays the groundwork for future enhancements. This could include community-driven efforts to create implementation and verification guidance built on top of the core requirement set.
```
elarlang, line 27: "The ASVS is not a static document" - it may send a wrong signal like ASVS is a living standard. I would like to use something I used in my proposed version with the message, that "Security can not be never ready, so can not ASVS be never ready"

5.0/en/0x02-Preface.md
elarlang: Typo: "comprehensive viewo"

elarlang (changes requested):

5.0/en/0x02-Preface.md
tghosth: > 4 times "now" in this list

Not sure that is a problem if it is across multiple list items.

> "The standard is made to align more directly"

Maybe:

"This version of the standard has been designed to align more directly..." 

Or just leave it like it is.

> > Each requirement is now crafted to be self-explanatory, highlighting its rationale.

> It is a direction, can not state the statement is satisfied.

Suggest:

"Each requirement is now crafted to be self-explanatory, highlighting its rationale and why it exists."

5.0/en/0x02-Preface.md
tghosth: How about:
"L1 is designed as the initial step to adopting the ASVS, providing the first layer of defense."

5.0/en/0x02-Preface.md
tghosth: How about:

"Just as securing an application is never truly finished, neither is the ASVS."

5.0/en/0x02-Preface.md
elarlang: > Not sure that is a problem if it is across multiple list items.

I think the "now" should not be used at all. "now" is valid only _at the moment_, standard lives many years.

---

> This version of the standard has been designed to align more directly

Better than the current version.

---

> Each requirement is now crafted to be self-explanatory, highlighting its rationale and why it exists.

The statement is not valid, and each requirement does not have it. It is a provocative call to report all requirements that do not satisfy the statement. It is our direction, but it is not valid to state for every requirement.

And the "now" is not needed here as well.

tghosth: Ok what do you think of the changes here:
https://github.com/OWASP/ASVS/pull/3129/commits/91a5567b60462dbfa5f4bc786cada4c3abbead65

elarlang (approved):

PR #3130, opened 2025-05-21
Add a business logic TOCTOU requirement to resolve #3117

### LINKED ISSUE #3117: Create a L2 TOCTOU requirement
We used to have this in the business logic chapter. It got moved to concurrency but that requirement is a lot more low level so I think we still need  a more business logic level requirement as well, (see the second half of [this comment](https://github.com/OWASP/ASVS/issues/2960#issuecomment-2879045286).)

tghosth:
This Pull Request relates to issue #3117

Note that there are mapping fixes to do. I think that this should be mapped to the 4.0.3 TOCTOU requirement rather than the new requirement in 15.4.2.


elarlang:
I close this as I prepared the PR with mapping fixes.

PR #3137, opened 2025-05-21
Tweak wording about mlkem768x25519 and fix broken link (Firefox support)

randomstuff:
* Add name of the mechanism (mlkem768x25519)
* Fix broken link for Firefox support
* Change "group" → "key agreement method" because it's not really a group is it?


randomstuff:
Disclaimer: I don't know much about these PQ key agreements methods.

@elarlang @danielcuthbert @unprovable

elarlang:
Some style questions like "MAY" to lowercase, and I'm not fan of "and/or" usage in the document.

unprovable:
> Some style questions like "MAY" to lowercase, and I'm not fan of "and/or" usage in the document.

Strong disagree - it's RFC2119 language and it's very clear when you do it the sense in which you're using those words. Cf: https://datatracker.ietf.org/doc/html/rfc2119

elarlang:
> > Some style questions like "MAY" to lowercase, and I'm not fan of "and/or" usage in the document.
> 
> Strong disagree - it's RFC2119 language and it's very clear when you use it the sense in which you're using those words. Cf: https://datatracker.ietf.org/doc/html/rfc2119

I am aware of the RFC, but in the ASVS, we don't use MUST and SHOULD in requirements; this is the style used throughout the document.

elarlang (approved):

PR #3146, opened 2025-05-22
Style/grammar updates for chapter/section text to resolve #3013

### LINKED ISSUE #3013: Handle spelling and grammar
Spelling in the ASVS should be consistently English US. 

We probably want to make minor grammar fixes as well.

My plan was to use copilot and run each file with the following prompt:

"_Please review this page and change any word which has a non US spelling to be its US spelling instead. Please also fix any major grammer issues. Do not change anything else_"

This is obviously something for towards the end.

I will add two sub-issues which were specifically already noted and should definitely be fixed.

tghosth:
I am adding files as I check the Copilot output. I don't want to go through a major iterations process on this so please only query things that you think are major, breaking changes or are just plain wrong...

### DIFF
--- a/5.0/en/0x10-V1-Encoding-and-Sanitization.md
+++ b/5.0/en/0x10-V1-Encoding-and-Sanitization.md
@@ -2,15 +2,15 @@
 
 ## Control Objective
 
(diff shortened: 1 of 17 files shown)


elarlang:
Note that the references section for V8, the "also" was removed from "For more information, see also:". For other chapters, it is not removed.

elarlang (approved):

PR #3149, opened 2025-05-23
Glossary_Suggestions_for_remove_and_update_definition

oleksiidov:
This Pull Request relates to issue #2201
Changes:
- suggestion on deletion of unused terms
- slightly corrected style
- suggested on corrections for some terms.

PR #3165, opened 2025-05-28
Glossary_Suggested_terms

oleksiidov:
This Pull Request relates to issue #3149

Suggested security related terms mentioned in document but missed in Glossary. Also term "Time-based OTP" aligned to used in document "Time based One-time Passwords" (TOTPs)


oleksiidov (commented):
Corrected issues for this Pull Request: #2201

PR #3177, opened 2025-06-02
Change factorisation -> factorization (US spelling)

randomstuff:


tghosth:
@elarlang, we don't need to worry about change tracking on anything that isn't a requirement, right? 

We should probably also decide how we want to handle changes to requirements but that is for another time :)

elarlang:
> @elarlang, we don't need to worry about change tracking on anything that isn't a requirement, right?

No worries.

> We should probably also decide how we want to handle changes to requirements but that is for another time :)

Soon.

elarlang (approved):

PR #3185, opened 2025-06-03
Correct branch name....

tghosth:


elarlang:
Wasn't there a campaign that the word "master" is bad-bad-bad?

tghosth:
> Wasn't there a campaign that the word "master" is bad-bad-bad?

Yes but I've always been too scared to break things by changing the name...

Issue #3224, opened 2025-08-01
Clarification: 3.2.2 Incompatible with Modern SPA Framework Practices


ajayojha:
> Verify that content intended to be displayed as text, rather than rendered as
HTML, is handled using safe rendering functions (such as createTextNode or
textContent) to prevent unintended execution of content such as HTML or
JavaScript.

In Single Page Application frameworks like react, vue, angular, using `textContent` or `createTextNode` manually is uncommon and unnecessary unless doing manual DOM manipulation. These API violates the declarative nature of Modern SPA frameworks, where rendering is handled through templating systems or virtual DOM abstractions.  Forcing to use `textContent` discourages declarative UI and reusable components, and undecessary complexity. Modern SPA frameworks automatically escape untrusted data during rendering by default, safe rendering is already inherently addressed.

elarlang:
> such as createTextNode or textContent

It is given as an example

> Forcing to use textContent discourages declarative UI and reusable components, and unnecessary complexity.

How given **example** forcing something?

> Modern SPA frameworks automatically escape untrusted data during rendering by default, safe rendering is already inherently addressed.

Then you are done and there is no problem to solve.

ajayojha:
> How given example forcing something?


The term used in these SPA frameworks like “**interpolation**” or “**data binding**” instead of “**safe rendering functions**,”. but the requirement focuses on "safe rendering functions" which only exist in **vanilla JavaScript**. 

In modern SPA frameworks like React, Vue, and Angular, safe text rendering is achieved through declarative mechanisms such as JSX interpolation (React), text interpolation (Vue), or property binding (Angular), which automatically escape untrusted data. The term “**safe rendering functions**” is not typically used in SPAs, as these frameworks rely on templating or virtual DOM abstractions rather than low-level DOM APIs. 

The requirement’s terminology and examples could mislead developers or auditors into enforcing non-declarative APIs in SPA codebases.

> Then you are done and there is no problem to solve.

Is the Developers & Testers just need to ignore the requirement?

randomstuff:
> The requirement’s terminology and examples could mislead developers or auditors into enforcing non-declarative APIs in SPA codebases.

I think that the intent in the context of SPA frameworks is to disallow the usage of things like React `dangerouslySetInnerHTML`, Vue.js `x-html` and Angular `[innerHTML]` when you intend to render text. If this is not clear enough, maybe the wording could be clarified in this direction? (I personally do *not* claim that this is not clear enough.)

[...]

PR #3251, opened 2025-09-16
Update v3.2.2 requirement #3224

ajayojha:
This Pull Request relates to issue #3224


elarlang:
Converted to draft:
* we don't have an agreement on the leadership side how do we track changes (and there should be no changes to the requirements before that)
* the wording impact itself is questionable (this discussion should be in the issue)

ajayojha:
> Converted to draft:
> 
> * we don't have an agreement on the leadership side how do we track changes (and there should be no changes to the requirements before that)

It’s a good to create a separate ticket and specify how many items are pending a decision from leadership. This way, volunteers are aware of the pending items before creating a ticket or PR. The ticket should also be pinned.

PR #3289, opened 2025-11-11
Fix: Updated outdated threat modeling reference URL in ASVS 5.0

### LINKED ISSUE #3287: Fix threat modeling link
> Hi, I noticed that ASVS 5.0 points to https://owasp.org/www-community/Application_Threat_Modeling as the 2nd reference on page 118. Can I request that be updated to https://owasp.org/www-project-threat-model/ ?

From Adam Shostack here:
https://owasp.slack.com/archives/C06MNF14M/p1762695739592259

Shreyas-Madake:
Updated the outdated reference URL in all ASVS 5.0 language versions.

The previous link:
https://owasp.org/www-community/Application_Threat_Modeling

has been replaced with the correct link:
https://owasp.org/www-project-threat-model/

Updated files:
- 5.0/en/0x93-Appendix-D_Recommendations.md
- 5.0/tr/0x93-Appendix-D_Recommendations.md
- 5.0/ru/0x93-Appendix-D_Recommendations.md
- 5.0/fr/0x93-Appendix-D_Recommendations.md

Fixes #3287

### DIFF
--- a/5.0/fr/0x93-Appendix-D_Recommendations.md
+++ b/5.0/fr/0x93-Appendix-D_Recommendations.md
@@ -44,6 +44,6 @@ Plusieurs processus de sécurité ont été supprimés d'ASVS 5.0, mais restent 
 Références:
 
 * [OWASP Threat Modeling Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Threat_Modeling_Cheat_Sheet.html)
-* [OWASP Threat modeling](https://owasp.org/www-project-threat-model/)
+* [OWASP Threat modeling](https://owasp.org/www-community/Application_Threat_Modeling)
 * [OWASP Software Assurance Maturity Model Project](https://owasp.org/www-project-samm/)
 * [Microsoft SDL](https://www.microsoft.com/en-us/securityengineering/sdl/)
--- a/5.0/ru/0x93-Appendix-D_Recommendations.md
+++ b/5.0/ru/0x93-Appendix-D_Recommendations.md
@@ -44,6 +44,6 @@
 Ссылки:
 
 * [OWASP Threat Modeling Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Threat_Modeling_Cheat_Sheet.html)
-* [OWASP Threat modeling](https://owasp.org/www-project-threat-model/)
+* [OWASP Threat modeling](https://owasp.org/www-community/Application_Threat_Modeling)
 * [OWASP Software Assurance Maturity Model Project](https://owasp.org/www-project-samm/)
 * [Microsoft SDL](https://www.microsoft.com/en-us/securityengineering/sdl/)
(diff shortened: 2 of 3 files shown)


elarlang:
I took a quick view of those projects, and I'm not convinced why "the new one" is better than "the old one"?

From a wider point of view, are those projects duplicates, or is the second one effectively a new version of the previous one? In both cases, those projects should be merged then, and the redirection done on the web page side, instead of fixing the link for every usage on the Internet (which is not realistic anyway).

Based on repo, "the new" project seems also a bit outdated?

The second question I have is, should we actually fix it in translations, as translation must be exactly the released version as it was (even in the case the released version contains a mistake). Otherwise, we have translations of bleeding-edge version that has the number of the released version, and that is not correct.

Shreyas-Madake:
> I took a quick view of those projects, and I'm not convinced why "the new one" is better than "the old one"?
> 
> From a wider point of view, are those projects duplicates, or is the second one effectively a new version of the previous one? In both cases, those projects should be merged then, and the redirection done on the web page side, instead of fixing the link for every usage on the Internet (which is not realistic anyway).
> 
> Based on repo, "the new" project seems also a bit outdated?
> 
> The second question I have is, should we actually fix it in translations, as translation must be exactly the released version as it was (even in the case the released version contains a mistake). Otherwise, we have translations of bleeding-edge version that has the number of the released version, and that is not correct.

Thanks for the review.
If the two pages are the same or one replaces the other, they should be merged and the old URL should redirect on the OWASP site.
If the “new” page is outdated, we should pick one official page, update it, and redirect to it.

For translations: we must match the released English exactly. We shouldn’t change links only in translations. If English is updated, I’ll update the translation to match.

Shreyas-Madake:
> Updated the outdated reference URL in all ASVS 5.0 language versions.
> 
> The previous link: https://owasp.org/www-community/Application_Threat_Modeling
> 
> has been replaced with the correct link: https://owasp.org/www-project-threat-model/
> 
> Updated files:
> 
> * 5.0/en/0x93-Appendix-D_Recommendations.md
> * 5.0/tr/0x93-Appendix-D_Recommendations.md
> * 5.0/ru/0x93-Appendix-D_Recommendations.md
> * 5.0/fr/0x93-Appendix-D_Recommendations.md
> 
> Fixes #3287

Reverted tr/ru/fr to released content;
 kept English updated. Ready for review.

adamshostack:
> I took a quick view of those projects, and I'm not convinced why "the new one" is better than "the old one"?
> 
> From a wider point of view, are those projects duplicates, or is the second one effectively a new version of the previous one? In both cases, those projects should be merged then, and the redirection done on the web page side, instead of fixing the link for every usage on the Internet (which is not realistic anyway).

One is a community page, the other is a project page. That's visible in the URL. We'll likely fix the underlying problem, but in the meantime I've asked that OWASP, in its official projects, link to other projects, rather than community pages.

tghosth (approved):

Issue #3322, opened 2026-01-05
Clarify CORS validation in 3.4.2 to explicitly reject 'null' origin


Galaxy-sc:
### Requirement Referenced
[3.4.2](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x12-V3-Web-Frontend-Security.md) - Browser Security Mechanism Headers

### Description of the Issue
The current text for requirement **3.4.2** mandates validating the `Origin` header against an allowlist but does not explicitly warn against accepting the `null` origin.

The `null` origin is a specific edge case generated by browsers in scenarios such as sandboxed iframes or local file access. If an application's allowlist is not strict enough (or if it reflects the origin blindly), an attacker can use a sandboxed iframe to bypass CORS policies by sending `Origin: null`.

### Proposed Change
I propose updating the requirement text to explicitly mention rejecting `null` unless it is strictly required by the architecture.

**Current Text:**
> ...validated against an allowlist of trusted origins. When 'Access-Control-Allow-Origin: *' needs to be used...

**Proposed Text:**
> ...validated against an allowlist of trusted origins **(specifically rejecting the `null` origin unless strictly required)**. When 'Access-Control-Allow-Origin: *' needs to be used...

### Motivation
This clarification ensures that developers and security testers specifically verify that the application does not inadvertently allow the `null` origin, closing a common bypass vector for CORS protections.

elarlang:
The requirement 3.4.2 as it is at the moment:

| # | Description | Level |
| :---: | :--- | :---: |
| **3.4.2** | Verify that the Cross-Origin Resource Sharing (CORS) Access-Control-Allow-Origin header field is a fixed value by the application, or if the Origin HTTP request header field value is used, it is validated against an allowlist of trusted origins. When 'Access-Control-Allow-Origin: *' needs to be used, verify that the response does not include any sensitive information. | 1 |

Well, I agree that the `null` option is not specially highlighted, but...

> The current text for requirement 3.4.2 mandates validating the Origin header against an allowlist but does not explicitly warn against accepting the null origin.

If the `null` is not "your fixed value" or in your allow list, it is not allowed anyway.


> The null origin is a specific edge case generated by browsers in scenarios such as sandboxed iframes or local file access. If an application's allowlist is not strict enough (or **if it reflects the origin blindly**), an attacker can use a sandboxed iframe to bypass CORS policies by sending Origin: null.

[...]

PR #3323, opened 2026-01-05
docs(3.4.2): clarify CORS validation to explicitly reject 'null' origin

Galaxy-sc:
This Pull Request relates to issue #3322

### Summary of Changes
Updated requirement **3.4.2** to explicitly specify that the `null` origin must be rejected when validating the `Origin` header, unless strictly required.

### Motivation
The `null` origin is a critical edge case in CORS configurations that is often overlooked. Attackers can trigger a `null` origin using sandboxed iframes or local file contexts to bypass CORS restrictions if the server carelessly reflects the `Origin` header or whitelists `null`.

Explicitly adding this check to the standard ensures developers and testers are aware of this vector and validates against it, closing a common loophole in "Reflected Origin" implementations.


elarlang:
marked as draft
* as there is no agreement in for the change in the issue yet
* requires modification tags
* requires formating change (no code block style allowed in the requirement)

Galaxy-sc:
I have corrected the formatting issues (removed code blocks). Regarding the 'modification tags', I checked the contribution guidelines but I am not sure which specific tag format is required for ASVS 5.0 changes. Could you please point me to the correct tag syntax or an example?

Also, I have replied to your concerns in the associated Issue regarding the necessity of this change.

elarlang:
> Could you please point me to the correct tag syntax or an example?

This is something that one of the leaders needs to keep an eye on. We have not finalized/updated the post-v5.0 procedure, and this may hold changes back a bit (to be applied to the repository, changes can be agreed anyway).

Galaxy-sc:
Understood. Thanks for the clarification.

I will leave the PR as is regarding the tags, pending the finalization of the procedure. In the meantime, I look forward to reaching an agreement on the actual content change in the linked Issue.

PR #3342, opened 2026-03-02
Retry status-0 URL check failures with curl

tghosth:
## Summary
- Increase `markdown-link-check` timeout from default 10s to 30s
- When `markdown-link-check` reports status-0 failures (connection refused/timeout), retry those specific URLs with `curl` before failing the job
- Genuine dead links (404, 500, etc.) still fail immediately without retry

## Context
The `cr.yp.to` URLs in the cryptography appendix consistently return status 0 from GitHub Actions runners despite being valid, reachable URLs. Rather than ignoring these URLs (which defeats the purpose of link checking), this adds a curl-based retry so that connection-level failures get a second chance while HTTP-level failures are still caught.

## Test plan
- [x] Triggered workflow on this branch — [run passed](https://github.com/OWASP/ASVS/actions/runs/22570845220)
- [x] Confirmed curl retry correctly identifies all 4 `cr.yp.to` URLs as alive (200)
- [x] Verified `set -o pipefail` correctly propagates `markdown-link-check` exit codes through `tee`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/url-checker.yml
+++ b/.github/workflows/url-checker.yml
@@ -37,6 +37,8 @@ jobs:
           cat /tmp/files_to_check.txt
 
       - name: Check URLs
+        id: linkcheck
+        continue-on-error: true
         run: |
           set -o pipefail
           ERROR_FOUND=0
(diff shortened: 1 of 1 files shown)


tghosth:
@elarlang

Issue #3344, opened 2026-03-05
V4.3.3-derived requirement accidentally deleted, mapping is stale


tghosth:
## Background

The mapping file shows `v5.0.0-13.2.4: MODIFIED, MOVED FROM v4.0.3-4.3.3`, but V13.2.4's current text has no relationship to V4.3.3. The V4.3.3-derived requirement was accidentally overwritten in April 2025 and no longer exists anywhere in the standard.

The last version of the requirement was:

> Verify that, if the application allows changing configurations around credentials or connection parameters for integrations with databases and services, they are protected by extra controls such as authenticating again with at least one factor or multi-user approval.

## How it was lost

The requirement originated from a long discussion in #1437 about protecting service integration configuration from being modified by attackers who gain admin access (e.g., via XSS or session hijacking). The original [V4.3.3](https://github.com/OWASP/ASVS/blob/v4.0.3/4.0/en/0x12-V4-Access-Control.md#v43-other-access-control-considerations) was about step-up authentication and anti-fraud controls. It was rewritten in #1437 to focus on config-change protection, moved to V14.7.3, and renumbered to V13.2.4 during the v5.0.0 restructuring.

In #2951, four changes to V13.2 were proposed:

- **Point 2** proposed minor wording fixes to the V4.3.3-derived text (passwords → credentials, remove "external"). This was applied in commit [`935c0292`](https://github.com/OWASP/ASVS/commit/935c0292), which also reordered the section so the updated text moved from 13.2.3 to 13.2.4.
- **Point 3** proposed rewriting the V12.6.1-derived text (which moved from 13.2.4 to 13.2.5 in the same reordering commit) to be technology-agnostic. The discussion produced new wording agreed by @elarlang, @jmanico, and @tghosth.

Three days later, PR #2982 applied Point 3's new wording to the **13.2.4 slot** — overwriting the Point 2 text that had just been updated. The commit message ([`e4c9d291`](https://github.com/OWASP/ASVS/commit/e4c9d291)) says "update 13.2.4 / v5.0.be-14.7.3, closes #2951". Nobody in the issue discussed deleting the V4.3.3-derived requirement — the intent was to rewrite V13.2.5 (the 12.6.1 text), not to replace V13.2.4.

## Questions

1. **Should the V4.3.3-derived requirement be restored?** It addressed a real security concern (protecting service credential/connection configuration from being modified via the application's admin interface) that was discussed at length in #1437 and is not covered by any other current requirement.

[...]

Issue #3343, opened 2026-03-05
Are 13.2.4 and 13.2.5 duplicates?


tghosth:
## Background

V13.2.4 and V13.2.5 sit next to each other in the Backend Communication Configuration section and both require allowlisting external communications:

| # | Description | Level |
| :---: | :--- | :---: |
| **13.2.4** | Verify that an allowlist is used to define the external resources or systems with which the application is permitted to communicate (e.g., for outbound requests, data loads, or file access). This allowlist can be implemented at the application layer, web server, firewall, or a combination of different layers. | 2 |
| **13.2.5** | Verify that the web or application server is configured with an allowlist of resources or systems to which the server can send requests or load data or files from. | 2 |

V13.2.5 appears to be a specific instance of V13.2.4 — if you satisfy V13.2.4 by configuring an allowlist at the web server level, you've also satisfied V13.2.5. Is there a reason to keep both?

## How we got here

**V13.2.5** is a direct descendant of [V12.6.1](https://github.com/OWASP/ASVS/blob/v4.0.3/4.0/en/0x20-V12-Files-Resources.md#v126-ssrf-protection) in v4.0.3:

> Verify that the web or application server is configured with an allow list of resources or systems to which the server can send requests or load data/files from.

The wording changed only minimally ("allow list" → "allowlist"). It was moved from V12 to V14 in #1324/#1343, then renumbered through the v5.0.0 pipeline (V14.7.1 → V14.6.1 → V14.7.4 → V13.2.5).

**V13.2.4** has a more complex history. Its *requirement slot* traces back to [V4.3.3](https://github.com/OWASP/ASVS/blob/v4.0.3/4.0/en/0x12-V4-Access-Control.md#v43-other-access-control-considerations) in v4.0.3, which was about step-up authentication and anti-fraud controls. That requirement was modified in #1437 to focus on protecting configuration changes for service credentials, moved to V14.7.3, and renumbered to V13.2.4. However, in #2951, @elarlang noted that the old V12.6.1 wording (then at V13.2.5) was too narrow — requiring configuration specifically in the "web or application server" when in practice the allowlist might be enforced by a WAF, firewall, application code, or a combination. @jmanico and @tghosth agreed and collaboratively drafted broader, technology-agnostic wording.

[...]

PR #3373, opened 2026-07-18
Fix stale v4.0.3-4.3.3 <-> v5.0.0-13.2.4 mapping

Sofeso:
Related to #3344 (does not close it — see "Scope" below).

## Problem

`mapping_v5.0.0_to_v4.0.3.yml` and `mapping_v4.0.3_to_v5.0.0.yml` both claim `v5.0.0-13.2.4` is a modified/moved version of `v4.0.3-4.3.3`. That's no longer accurate:

- `v4.0.3-4.3.3` was about requiring step-up auth / multi-user approval before changing credential or connection-parameter configuration for database/service integrations.
- Current `v5.0.0-13.2.4` reads: *"Verify that an allowlist is used to define the external resources or systems with which the application is permitted to communicate..."* — a completely different topic (application-layer allowlisting, the counterpart to 13.2.5's server-layer allowlisting).

## Independent verification

I re-derived the facts in #3344 myself rather than taking the issue at face value:

1. **Confirmed 13.2.5's lineage is intact**: `4.0/en/0x20-V12-Files-Resources.md` V12.6.1 ("Verify that the web or application server is configured with an allow list of resources or systems to which the server can send requests or load data/files from.") matches current V13.2.5 almost word-for-word — consistent with 13.2.5's own tag (`GRAMMAR, MOVED FROM v4.0.3-12.6.1`).
2. **Confirmed 13.2.4 has no v4.0.3 lineage at all**: searched all of `4.0/en/*.md` for wording resembling 13.2.4's actual text (allowlist + external + communicate) — no match anywhere. It's genuinely new content in v5.0.0, not a "modified" version of anything.
3. **Confirmed exactly where the mapping went stale**: the intermediate bleeding-edge files (`mapping_v4.0.3_to_v5.0.be.yml`, `mapping_v5.0.be_to_v4.0.3.yml`) correctly show `v4.0.3-4.3.3 -> v5.0.be-14.7.3`, and `mapping_v5.0.be_to_v5.0.0.yml` correctly renumbers `v5.0.be-14.7.3 -> v5.0.0-13.2.4`. That renumbering was accurate *at the time*. It's only the two final `v4.0.3<->v5.0.0` mapping files that are stale, because — per the commit history #3344 lays out — `13.2.4`'s *text* (not its ID) was later overwritten by unrelated 12.6.1-derived content. So I only touched those two files; the "be" intermediate mappings still correctly describe history and don't need changes.

## Change

```diff
 v4.0.3-4.3.3:
-  tag-v5.0.0: MOVED TO v5.0.0-13.2.4
+  tag-v5.0.0: DELETED
```
```diff
 v5.0.0-13.2.4:
-  tag-v4.0.3: MODIFIED, MOVED FROM v4.0.3-4.3.3
+  tag-v4.0.3: ADDED
```

## Scope — what this PR intentionally does NOT do

#3344 also asks whether the lost 4.3.3-derived control (step-up auth for credential/connection config changes) should be *restored* as a new requirement. That's a substantive content/policy decision for ASVS maintainers, not something I'm deciding via this PR. This PR only fixes the factual bookkeeping error so the mapping accurately reflects what the standard says today. I've left #3344 open rather than marking it closed, since that question is still unresolved.

## Disclosure

I used AI assistance (Claude) to investigate this issue and draft the fix, including independently re-verifying the git-archaeology claims in #3344 against the actual file contents rather than trusting them directly. I reviewed the evidence and take responsibility for the change.


elarlang:
Thank you for digging into it, but for approving it, I need to re-analyze issues #3343 and #3344 and it may take some (calendar) time to happen.

